Issue #14, opened 2014-05-06
More input checking needed on configuration files


campagnola:
Writing configuration files is one of the major barriers to getting ACQ4 running; we would benefit greatly from thorough input checking for all configuration parameters to provide helpful error messages at startup.

Ref: https://groups.google.com/d/msgid/acq4/32d42d80-dce4-4f99-b974-699f1ea19062%40googlegroups.com?utm_medium=email&utm_source=footer

PR #15, opened 2014-05-07
Parameters of Patch modules can be changed in the config file

ablot:
Following your advice (https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/acq4/fxn-E2YTkt0/sJoz2U_O6tIJ) I made a PR.

Reading the config is indeed relatively straight forward. I've added a quick and clumsy "type checking" so that it displays some errors, but I'm not convinced it is what you where suggesting in https://github.com/acq4/acq4/issues/14

Anyway it can easily be removed for now.


campagnola:
FYI I am working on this but have not figured out how to get it working on both of our systems at the same time. Need to determine the correct logic for interrogating which boards / capabilities are present and deciding on the correct click/trigger configuration.

PR #22, opened 2015-05-12
Added the possibility to chose between 200A or 200B AxoPatch amplifier.

mgraupe:
0.1 version.

### DIFF
--- a/acq4/devices/AxoPatch200/AxoPatch200.py
+++ b/acq4/devices/AxoPatch200/AxoPatch200.py
@@ -12,31 +12,22 @@ from collections import OrderedDict
 from acq4.util.debug import printExc
 from devGuiTemplate import *
 
-# telegraph voltage/output translation from the Axopatch 200 amplifier
-# modfied parameter for AxonPatch 200A
-#mode_tel = np.array([6, 4, 3, 2, 1])
-#mode_tel = np.array([6, 4, 2, 1])
-#mode_char = ['V', 'T', '0', 'I', 'F']
-#modeNames = OrderedDict([(0, 'V-Clamp'), (2, 'I=0'), (4, 'I-Clamp Fast'), (3, 'I-Clamp Normal'), (1, 'Track'), ])
-#modeNames = OrderedDict([(0, 'V-Clamp'), (1, 'Track'), (2, 'I-Clamp Normal'), (3, 'I-Clamp Fast'),  ])
-#ivModes = {'V-Clamp':'vc', 'Track':'vc', 'I=0':'ic', 'I-Clamp Fast':'ic', 'I-Clamp Normal':'ic', 'vc':'vc', 'ic':'ic'}
-#ivModes = {'V-Clamp':'vc', 'Track':'vc', 'I-Clamp Fast':'ic', 'I-Clamp Normal':'ic', 'vc':'vc', 'ic':'ic'}
-#modeAliases = {'ic': 'I-Clamp Fast', 'i=0': 'Track', 'vc': 'V-Clamp'}
 
-# Axopatch gain telegraph
-# telegraph should not read below 2 V in CC mode
-#gain_tel = np.array([0.5,  1.0, 1.5, 2.0, 2.5, 3.0, 3.5, 4.0, 4.5, 5.0, 5.5, 6.0, 6.5])
-#gain_vm  = np.array([0.5,  0.5, 0.5, 0.5, 1,   2,   5,   10,  20,  50,  100, 200, 500]) * 1e9  ## values in mv/pA
-#gain_im  = np.array([0.05, 0.1, 0.2, 0.5, 1,   2,   5,   10,  20,  50,  100, 200, 500])        ## values in mV/mV
-
-# Axopatch LPF telegraph
-#lpf_tel = np.array([2.0, 4.0, 6.0, 8.0, 10.0])
-#lpf_freq = np.array([1.0, 2.0, 5.0, 10.0, 100.0])
 
         #GainChannel: 'DAQ', '/Dev1/ai14'
         #LPFchannel: 'DAQ', '/Dev1/ai15'
         #VCommand: 'DAQ', '/Dev1/ao0'
         #ScaledSignal: 'DAQ', '/Dev1/ai5'
+try:
+    _encoding = QtGui.QApplication.UnicodeUTF8
+    def _translate(context, text, disambig):
+        return QtGui.QApplication.translate(context, text, disambig, _encoding)
+except AttributeError:
+    def _translate(context, text, disambig):
+        return QtGui.QApplication.translate(context, text, disambig)
+
+
+
 class AP200DataMapping(DataMapping):
     def __init__(self, dev, ivModes, chans=None, mode=None):
         ## mode can be provided:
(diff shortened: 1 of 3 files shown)


acq4/devices/AxoPatch200/AxoPatch200.py
```diff
@@ -126,6 +131,43 @@ def __init__(self, dm, config, name):
         else:
             self.hasSecondaryChannel = False
         
+        if 'AxoPatchVersion' in config :
+			if config['AxoPatchVersion'] == '200A':
```
campagnola: Replace tabs with 4 spaces. Most code editors can be configured to do this by default.
See: 
https://www.python.org/dev/peps/pep-0008/#id12
https://www.python.org/dev/peps/pep-0008/#tabs-or-spaces

acq4/devices/AxoPatch200/devGuiTemplate.py
```diff
@@ -24,7 +24,8 @@ def _translate(context, text, disambig):
         return QtGui.QApplication.translate(context, text, disambig)
 
 class Ui_devGui(object):
-    def setupUi(self, devGui):
```
campagnola: This file is auto-generated from `devGuiTemplate.ui`  (as the header indicates, changes to this file will be overwritten the next time it is regenerated). Instead, I would move the `modeCombo.setItemText` calls to `AxoPatch200.py`, and use Qt Designer to remove the combo box items from the `.ui` file. Finally, run `pyuic4 devGuiTemplate.ui > devGuiTemplate.py` to regenerate the python file.

acq4/devices/AxoPatch200/AxoPatch200.py
```diff
@@ -126,6 +131,43 @@ def __init__(self, dm, config, name):
...
+				self.gain_im  = np.array([0.05, 0.1, 0.2, 0.5, 1,   2,   5,   10,  20,  50,  100, 200, 500])        ## values in mV/mV
+
+				# Axopatch LPF telegraph
+				self.lpf_tel = np.array([2.0, 4.0, 6.0, 8.0, 10.0])
+				self.lpf_freq = np.array([1.0, 2.0, 5.0, 10.0, 100.0])
+
+			elif config['AxoPatchVersion'] == '200B':
+				# telegraph voltage/output translation from the Axopatch 200 amplifier
+				self.mode_tel = np.array([6, 4, 3, 2, 1])
+				#self.modeNames = OrderedDict([(0, 'V-Clamp'), (2, 'I=0'), (4, 'I-Clamp Fast'), (3, 'I-Clamp Normal'), (1, 'Track'), ])
```
campagnola: Why is this commented?

acq4/devices/AxoPatch200/AxoPatch200.py
```diff
@@ -126,6 +131,43 @@ def __init__(self, dm, config, name):
...
+				self.ivModes = {'V-Clamp':'vc', 'Track':'vc', 'I=0':'ic', 'I-Clamp Fast':'ic', 'I-Clamp Normal':'ic', 'vc':'vc', 'ic':'ic'}
+				self.modeAliases = {'ic': 'I-Clamp Fast', 'i=0': 'I=0', 'vc': 'V-Clamp'}
+				self.requiredSwitchMode = 'I=0'
+				# Axopatch gain telegraph
```
campagnola: Gain and lpf data appear to be identical between the versions. If this is correct, then they can be defined just once outside the `if` block.

campagnola:
Thanks, this looks good! The approach to use 'AxoPatchVersion' in the config file and get rid of the global variables is exactly what I would have suggested.

acq4/devices/AxoPatch200/AxoPatch200.py
```diff
@@ -13,28 +13,32 @@
 from devGuiTemplate import *
 
 # telegraph voltage/output translation from the Axopatch 200 amplifier
-mode_tel = np.array([6, 4, 3, 2, 1])
+# modfied parameter for AxonPatch 200A
```
campagnola: You can delete these rather than commenting them.

mgraupe: OK. No problem.

acq4/devices/AxoPatch200/AxoPatch200.py
mgraupe: OK.

acq4/devices/AxoPatch200/AxoPatch200.py
mgraupe: Oups. Sorry. Will uncomment it.

acq4/devices/AxoPatch200/AxoPatch200.py
mgraupe: The lpf_tel is the same but the lpf_freq differ between the models. The highest frequency is 50 kHz on the 200A while 100 kHz on the 200B. 
I moved lpf_tel outside the if block.

mgraupe: Btw, why does the gain_vm and gain_im variables have 13 values in the previous setting? There are 10 settings (0.5-500) on the gain dial with the 200A and B.

campagnola: If I remember correctly, these were the values specified in the manual we have. I also seem to recall there are some "unofficial" versions of the AxoPatch; perhaps some of these have different settings..

acq4/devices/AxoPatch200/devGuiTemplate.py
mgraupe: Oh, sure. Sorry.

I did as you suggested but the modeCombo options don't appear in the central gui once acq4 is launched. 

I also had to copy these definitions in the AxoPatch200.py file which I doubt is ideal. 

try:
    _encoding = QtGui.QApplication.UnicodeUTF8
    def _translate(context, text, disambig):
        return QtGui.QApplication.translate(context, text, disambig, _encoding)
except AttributeError:
    def _translate(context, text, disambig):
        return QtGui.QApplication.translate(context, text, disambig)

Do you have an idea why it is not working.

acq4/devices/AxoPatch200/AxoPatch200.py
```diff
@@ -561,13 +594,24 @@ def getChanHolding(self, chan):
...
-    def __init__(self, dev):
+    def __init__(self, dev,AxoPatchVersion):
         QtGui.QWidget.__init__(self)
         self.dev = dev
         self.ui = Ui_devGui()
         self.ui.setupUi(self)
         self.ui.vcHoldingSpin.setOpts(step=1, minStep=1e-3, dec=True, suffix='V', siPrefix=True)
         self.ui.icHoldingSpin.setOpts(step=1, minStep=1e-12, dec=True, suffix='A', siPrefix=True)
+        if AxoPatchVersion == '200A':
+            self.ui.modeCombo.setItemText(0, _translate("devGui", "V-Clamp", None))
```
campagnola: You can replace these with `self.ui.modeCombo.setItemText(0, "V-Clamp")`, which will allow you to remove the `_translate` definition above (we're only operating in one language anyway).

campagnola: Ah, actually try self.ui.modeeCombo.addItem("V-Clamp")` instead; that's probably why your list was coming up empty.

mgraupe: Of course, good point. The item hasn't been added at all. I followed you suggestion and it's working like a charm now... .

acq4/devices/AxoPatch200/AxoPatch200.py
campagnola: It is in the manual:
https://www.autom8.com/docs/Axopatch_200B.pdf

The total output gain is the product of two factors alpha * beta, where alpha is determined by the gain knob and beta is determined by the adjacent "config." switch. This appears to be true on the 200, so I presume it is on the 200A as well.

campagnola:
Thanks very much! I merged this with some changes:

* The config parameter that specifies the hardware version is renamed from `AxoPatchVersion` to `version`. 
* Updated the example `devices.cfg`
* Added support for AxoPatch 200 (the original)
* Corrected gain modes for 200A -- I think these should indeed go all the way down to X0.05 (but please let me know if this is not correct)
* Other minor cleanups

PR #23, opened 2015-05-19
fixes for the MapImager module against changes to pyqtgraph's ComboBox

meganbkratz:


campagnola:
I guess `updateList` used to accept tuples but doesn't anymore? (If so, do you think you can fix it to accept any sequence?)

meganbkratz:
ComboBox.addItems can now accept lists, dicts and tuples. Is there anything else it should accept?

It changed with this commit (looks like a big merge of pyqtgraph into acq4): 86dcb456d2c37cab07a62f86dd15424b6302179d. Before that we used to just pass items directly to Qt's QComboBox.addItems.

acq4/pyqtgraph/widgets/ComboBox.py
```diff
@@ -190,13 +190,13 @@ def setItemValue(self, name, value):
     @ignoreIndexChange
     @blockIfUnchanged
     def addItems(self, items):
-        if isinstance(items, list):
+        if isinstance(items, list) or isinstance(items, tuple):
```
campagnola: FYI: `isinstance(items, (list, tuple))`

PR #24, opened 2015-06-20
Save rect scan in task runner

campagnola:
Needs testing before merge.

### DIFF
--- a/acq4/devices/Scanner/TaskGui.py
+++ b/acq4/devices/Scanner/TaskGui.py
@@ -278,18 +278,23 @@ class ScannerTaskGui(TaskGui):
             return None
         return mod
         
+
     def saveState(self, saveItems=False):
         state = self.stateGroup.state()
         if saveItems:
             state['items'] = [param.item.saveState() for param in self.positionCtrlGroup.children()]
+        if self.ui.enableScanProgCheck.isChecked():
+            state['program'] = self.scanProgram.saveState()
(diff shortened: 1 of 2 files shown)


mgraupe:
Indeed, the changes allow to load a previously saved "Scan Program" in a Task Runner Protocol. All parameters are faithfully recovered upon loading of a protocol. 

However, a scan finishes with a TypeError (see below) and the scan ouput is not shown in the Imaging or the Camera tab. 

------------
TypeError: constraint must be None, True, 'fixed', or tuple. (got False)
  File "acq4\modules\TaskRunner\analysisModules\Imaging\Imaging.py", line 101, in newFrame
    self.update()  # updates image
  File "acq4\modules\TaskRunner\analysisModules\Imaging\Imaging.py", line 168, in update
    rs.restoreState(prog['scanInfo'])
  File "acq4\pyqtgraph\parametertree\SystemSolver.py", line 244, in restoreState
    self.set(name, var[0], var[1])
  File "acq4\pyqtgraph\parametertree\SystemSolver.py", line 195, in set
    raise TypeError("constraint must be None, True, 'fixed', or tuple. (got %s)" % constraint)
--------------

The set function in SystemSolver does not accept "False" for the scan program parameters which are not fixed. 

Cheers,
Michael

campagnola:
Ok, I believe this is fixed.

PR #27, opened 2015-07-06
Three changes in Multiphoton Imager module

mgraupe:
1. State of the window is saved when closing and recovered when re-opening. 2. defaultFieldSize can be configured in the Imager Module configuration. 3. The cameraModule window initializes to include Camera and Imager roi.


mgraupe:
Please note that the recent commit (7a84f48) contains already the changes made in eb2f8d0 .

acq4/modules/Camera/CameraWindow.py
```diff
@@ -189,12 +189,17 @@ def getDepthView(self):
         self.depthDock.show()
         return self.depthPlot
 
-    def centerView(self):
+    def centerView(self,exclusive=None):
```
campagnola: A docstring would be nice here explaining what the new argument is for.

PR #34, opened 2015-11-24
Scanner laser control

mgraupe:
This is related to PR #33 . 

Added signals to differentiate between single task run and a task sequence. The shutter will open or close between runs depending on releaseAfterSequence variable.


mgraupe:
The laser-mask implementation works great. I have been testing it extensively. 

The only part which hasn't been implemented yet is the simple-mode user interface for `getArray` that lists the available data sources and allows to set the pockel cell voltage. 

That is what Luke suggested but I am stuck with the implementation and could need some help:
"Should be easy indeed--just add another parameter class similar to PulseParameter that reads the key names from acq4.generator.DATA_SOURCES.keys(). You'll also need to inform the parameter whenever the list changes (see StimGenerator.dataSourceChanged)."

PR #44, opened 2016-03-24
Maitai laser

mgraupe:
This driver and device class adds control of the Spectra-Physics Mai Tai laser to ACQ4.


acq4/drivers/MaiTai/maitai.py
```diff
@@ -0,0 +1,280 @@
...
+    from ..SerialDevice import SerialDevice, TimeoutError, DataError
+except ValueError:
+    ## relative imports not allowed when running from command prompt, so
+    ## we adjust sys.path when running the script for testing
+    if __name__ == '__main__':
+        import sys, os
+        sys.path.append(os.path.abspath(os.path.join(os.path.dirname(__file__), '..')))
+        from SerialDevice import SerialDevice, TimeoutError, DataError
+        
+ErrorVals = {
```
campagnola: Probably don't need these here.

acq4/devices/MaiTaiLaser/MaiTaiLaser.py
```diff
@@ -0,0 +1,255 @@
...
+    
+    def acitvateAlignmentMode(self):
+        with self.driverLock:
+            self.greenPowerIRMode = self.driver.getLastCommandedPumpLaserPower()
+        with self.driverLock:
+            self.driver.setPumpMode('Green Power')
+        self.mThread.alignmentMode = True
+        
+    def deactivateAlignmentMode(self):
+        with self.driverLock:
```
campagnola: Do you need two independent driver locks?

        with self.driverLock:
            self.driver.setPumpLaserPower(self.greenPowerIRMode)
            self.driver.setPumpMode('IR Power')

acq4/devices/Laser/MaiTaiDevGui.py
```diff
@@ -0,0 +1,427 @@
+from PyQt4 import QtGui, QtCore
```
campagnola: Shouldn't this file be in /devices/MaiTaiLaser/ ?

acq4/drivers/MaiTai/maitai.py
```diff
@@ -0,0 +1,280 @@
...
+
+class TimeoutError(Exception):
+    pass
+
+class MaiTaiError(Exception):
+    pass
+
+class MaiTai(SerialDevice):
+    """
+    Class for communicating with Sutter MP-285 via serial port.
```
campagnola: Needs update

acq4/drivers/MaiTai/maitai.py
```diff
@@ -0,0 +1,280 @@
...
+    Note that this class is NOT thread-safe.
+    """
+    def __init__(self, port, baud=9600):
+        """
+        port: serial COM port (0 => com1)"""
+
+        self.re_float = re.compile(r'\d*\.?\d+')
+        self.port = port
+        self.baud = baud
+        self.sp = serial.Serial(int(self.port), baudrate=self.baud, bytesize=serial.EIGHTBITS,parity=serial.PARITY_NONE,stopbits=serial.STOPBITS_ONE,xonxoff=True)
```
campagnola: Your driver class inherits from SerialDevice but then creates its own second serial port handle.. See the Scientifica driver for an example of a serial device done "correctly".

acq4/drivers/MaiTai/maitai.py
```diff
@@ -0,0 +1,280 @@
...
+
+    def clearBuffer(self):
+        d = self.read()
+        time.sleep(0.1)
+        d += self.read()
+        if len(d) > 0:
+            print "Sutter MP285: Warning: tossed data ", repr(d)
+        return d
+    
+    def read(self):
```
campagnola: If you use SerialDevice correctly, then it probably won't be necessary to reimplement read / write / close / readPacket.  (and you can use readUntil instead of readPacket)

acq4/devices/Laser/devTemplate.py
```diff
@@ -26,104 +25,42 @@ def _translate(context, text, disambig):
...
-        self.gridLayout_5.addWidget(self.GDDEnableCheck, 1, 0, 1, 1)
-        self.GDDSpin = QtGui.QSpinBox(self.wavelengthGroup)
-        self.GDDSpin.setMaximum(32000)
-        self.GDDSpin.setObjectName(_fromUtf8("GDDSpin"))
-        self.gridLayout_5.addWidget(self.GDDSpin, 1, 1, 1, 1)
-        self.GDDLimits = QtGui.QLabel(self.wavelengthGroup)
-        self.GDDLimits.setObjectName(_fromUtf8("GDDLimits"))
-        self.gridLayout_5.addWidget(self.GDDLimits, 1, 2, 1, 1)
-        self.gridLayout.addWidget(self.wavelengthGroup, 3, 0, 1, 3)
+        self.maitaiControlGroup = QtGui.QGroupBox(Form)
```
campagnola: something is not right here.. the base Laser gui should not contain any information about maitai. It probably doesn't need to be changed very much. The main differences I see are:

* Added maitai-specific status display
* Modified wavelength control without GDD

For the first, we should be able to find a way to add the status display widget via a subclass, rather than by modifying the base gui. For the second, we can have the base laser class just hide the GDD controls if the laser does not support it.

mgraupe: I addressed all your feedback and will test it tomorrow morning. 

However, I am not sure how to create a dev gui subclass. My thinking was that the MaiTai control widgets can be hidden for any other laser model. That is why I added the modifications into the base laser gui. 

How can a dev gui subclass be implemented?

acq4/drivers/MaiTai/maitai.py
mgraupe: I removed read, write and close. readUntil gave errors as the response from the laser have inconsistent format and do not always end with the <cr> : '\n' . Some of the commands simply return a number. I tried to fix that but did not manage to get ride of the errors at this point. 

I therefor sticked with with readPacket().

acq4/devices/Laser/devTemplate.py
campagnola: Here's a plan:
- change the top-level layout of the base laser dev gui to a QVBoxLayout instead of QGridLayout (because the former has an insertWidget() method, and the latter does not).
- for the maitai dev gui, just create a widget containing the status info you want to show
- when you create the maitai dev gui, start with the base laser gui and then insert the extra widget wherever you like

acq4/drivers/MaiTai/maitai.py
campagnola: That's fine--if readUntil doesn't work for your device, then do whatever it takes. My main concerns are that 1) there is some confusion in this class because `SerialDevice.__init__` is never called and instead you manually open the serial port, and 2) code that is already working in SerialDevice should not be copied.

mgraupe: Of course, and I took your advice and use SerialDevice.__init__ now and removed the superfluous code.

acq4/devices/Laser/devTemplate.py
mgraupe: I implemented your suggestion as far as I could. I need however some help in inserting the Mai Tai widget in the laser dev gui:

- I changed the layout of the laser dev gui to QVBoxLayout. Please note that I placed the buttons in a "buttonGroup" which keeps their orientation and it allows to remove them in child laser classes. 
- I added all the Mai Tai specific control buttons and status information in a separate widget acq4/devices/MaiTaiLaser/maiTaiTemplate.ui 
- I am struggling however with the last point. I didn't manage to insert the widget in the base laser gui. I am sure this does not require much but I am running out of ideas how to do it. Please check out the latest version of this branch which contains all the changes.

Thank you for your help .

mgraupe: Great. Thanks. That did the job.

The Mai Tai Laser implementation feels ready now.

PR #45, opened 2016-05-02
Filterwheel

mgraupe:
The Filter wheel device class adds control and display for the filter wheel device status.


acq4/devices/FilterWheel/FilterWheel.py
```diff
@@ -0,0 +1,402 @@
+# -*- coding: utf-8 -*-
+from acq4.devices.OptomechDevice import *
```
campagnola: Usually preferred to avoid `import *`, since this makes it difficult to determine where symbols came from and can have unintended side-effects.

acq4/devices/FilterWheel/FilterWheel.py
```diff
@@ -0,0 +1,402 @@
...
+        
+        self.scopeDev = None
+        self.currentFilter = None
+        p = self
+        while p is not None:
+            p = p.parentDevice()
+            if isinstance(p, Microscope):
+                self.scopeDev = p
+                
+        self.port = config['port']-1  ## windows com ports start at COM1, pyserial ports start at 0
```
campagnola: Better to let the user specify port as a string like "com1" or "/dev/ttyS0". `SerialDevice` should take care of this for you.

acq4/devices/FilterWheel/FilterWheel.py
```diff
@@ -0,0 +1,402 @@
...
+            self.filters[k] = filt
+        
+        #print self.filters.values()
+        #print self.filters[1].name(), self.filters[1].description()
+        #if len(self.positionLabels) != self.getPositionCount():
+        #    raise Exception("Number of FilterWheel positions %s must correspond to number of labels!" % self.getPositionCount())
+        
+        with self.driverLock:
+            self.currentFWPosition = self.driver.getPos()
+            self.currentObjective = self.getFilter()
```
campagnola: Should this be `self.currentFilter` ?

acq4/devices/FilterWheel/FilterWheel.py
```diff
@@ -0,0 +1,402 @@
...
+            self.sigFilterWheelSpeedChanged.emit(speed)
+    
+    def getSpeed(self):
+        with self.driverLock:
+            return self.driver.getSpeed()
+        
+    def setPosition(self, pos):
+        with self.driverLock:
+            self.driver.setPos(pos)
+            self.sigFilterWheelPositionChanged.emit(pos)
```
campagnola: It seems like this will cause `sigFilterWheelPositionChanged` to be emitted twice--this one is emitted before the move, and then the thread will detect a position change and cause another to be emitted via `positionChanged`; is that correct?

acq4/devices/FilterWheel/FilterWheel.py
```diff
@@ -0,0 +1,402 @@
...
+        return self._fw
+    
+    def __repr__(self):
+        return "<Filter %s.%s>" % (self._fw.name(), self.name())
+
+
+    
+class FilterWheelTask(DeviceTask):
+    #pass
+    # This is disabled--internal shutter in coherent laser should NOT be used by ACQ4; use a separate shutter.
```
campagnola: remove

acq4/devices/FilterWheel/FilterWheel.py
```diff
@@ -0,0 +1,402 @@
...
+        self.cmd = cmd
+        self.parentTask = parentTask
+        #print parentTask
+        
+    def configure(self):
+        """Sets the state of a remote multiclamp to prepare for a program run."""
+        #print "mc configure"
+        with self.dev.filterWheelLock:
+            #self.state = self.dev.getLastState()
+            print 'command :', self.cmd
```
campagnola: remove

campagnola: (the print statements)

acq4/devices/FilterWheel/FilterWheel.py
```diff
@@ -0,0 +1,402 @@
...
+            requiredPos = int(self.cmd['filterWheelPosition'][0]) # that the first character of string and convert it to int
+            if self.dev.currentFWPosition != requiredPos:
+                print 'current pos:', self.dev.currentFWPosition
+                print 'requir. pos:', requiredPos
+                self.dev.setPosition(requiredPos)
+            
+    def start(self):
```
campagnola: Question: when you call `dev.setPosition`, does that method block until the filter change is finished, or does it return immediately? If it's the latter, then you probably need to block here until the filter is ready?

acq4/devices/FilterWheel/FilterWheel.py
```diff
@@ -0,0 +1,402 @@
...
+        self.stateGroup = pg.WidgetGroup([
+            (self.ui.filterCombo,),
+            (self.ui.sequenceCombo,),
+            (self.ui.sequenceListEdit,),
+        ])
+        
+        
+        
+    def generateTask(self, params=None):
+        print 'params in gernateTask :', params
```
campagnola: remove

acq4/devices/FilterWheel/FilterWheel.py
```diff
@@ -0,0 +1,402 @@
...
+        task['filterWheelPosition'] = target #state['filterCombo']
+        return task    
+    
+    def saveState(self, saveItems=False):
+        state = self.stateGroup.state()
+        return state
+    
+    def restoreState(self, state):
+        self.stateGroup.setState(state)
+        if state['sequenceCombo'] == 'off':
```
campagnola: This needs to hide and show as well, right? 

    self.ui.sequenceListEdit.setVisible(state['sequenceCombo'] != 'off')

acq4/drivers/ThorlabsFW102C/thorFW102cDriver.py
```diff
@@ -0,0 +1,189 @@
...
+from collections import OrderedDict
+
+errorMessages = {
+    'CMD_NOT_DEFINED': 0xEA,
+    'time out':0xEB,
+    'time out':0xEC,
+    'invalid string buffer':0XED
+    }
+
+class filterWheelDriver(SerialDevice):
```
campagnola: Capitalize class names

acq4/drivers/ThorlabsFW102C/thorFW102cDriver.py
```diff
@@ -0,0 +1,189 @@
...
+        SerialDevice.__init__(self, port=int(port), baudrate=baud)
+        
+    def getPos(self):
+        """Reads and returns the current filterwheel position """
+        pos = self['pos?']
+        return int(pos)
+    
+    def setPos(self, newPos):
+        """Sets filterwheel position to n """
+        print 'driver new pos ', newPos
```
campagnola: remove

acq4/drivers/ThorlabsFW102C/thorFW102cDriver.py
```diff
@@ -0,0 +1,189 @@
...
+class filterWheelDriver(SerialDevice):
+    """
+    Class for communicating with ThorLabs FW 102C Motorized Filter Wheel
+    
+    """
+    def __init__(self, port, baud=115200):
+        #self.fws = {}
+        #self.paramTable = OrderedDict()
+        
+        SerialDevice.__init__(self, port=int(port), baudrate=baud)
```
campagnola: Better to allow string ports rather than int.

acq4/modules/Imager/Imager.py
```diff
@@ -330,6 +330,10 @@ def __init__(self, manager, name, config):
         # to select from [(dev1, channel1), ...]
         self.detectors = config.get('detectors', [config.get('detector')])
         
+        self.filterwheel = self.manager.getDevice(config['filterwheel'])
```
campagnola: 1. Looks like this will fail if 'filterwheel' is not present in the config.
2. Maybe use 'filter' or 'filterDevice' instead of 'filterwheel', since we might like to support other types of filter-switching devices..

acq4/modules/TaskRunner/analysisModules/Imaging/Imaging.py
```diff
@@ -18,6 +19,8 @@ def __init__(self, *args):
         AnalysisModule.__init__(self, *args)
         # self.ui = Ui_Form()
         # self.ui.setupUi(self)
+        taskRunner = args
+        print 'taskRunner', dir(taskRunner)
```
campagnola: remove

acq4/modules/TaskRunner/analysisModules/Imaging/Imaging.py
```diff
@@ -37,13 +40,26 @@ def __init__(self, *args):
...
+            dict(name='Filter', type='str', value='Unknown', readonly=True),
             ])
         self.ptree.setParameters(self.params, showTop=False)
         self.params.sigTreeStateChanged.connect(self.update)
         self.params.child('detectors').sigAddNew.connect(self.addDetectorClicked)
         self.params.child('decomb', 'auto').sigActivated.connect(self.autoDecomb)
 
         self.man = getManager()
+        print self.man, dir(self.man.devices)
+        self.scannerDev = self.man.getDevice('Scanner')
```
campagnola: Don't assume 'Scanner'; the user selects the name of the scanner device.

acq4/modules/TaskRunner/analysisModules/Imaging/Imaging.py
```diff
@@ -37,13 +40,26 @@ def __init__(self, *args):
...
+        # find first scope device that is parent of scanner
+        dev = self.scannerDev
+        while dev is not None and not isinstance(dev, Microscope):
+            dev = dev.parentDevice()
+        self.scopeDev = dev
+        self.filterwheel = self.man.getDevice('FilterWheel')
```
campagnola: Obviously this won't work on systems without a device with this name.. I think the correct approach here would be something like:  1) make the PMT an optomech device and set up the hierarchy like Stage -> Microscope -> FilterWheel -> PMT;  2) ask the PMT whether it has a filter selection in its parent devices. To start, you could do this pretty simply:

    class PhotoDetector(DAQGeneric, OptomechDevice):
        def __init__(self, dm, config, name):
            omConf = {}
            for k in ['parentDevice', 'transform']:
                if k in config:
                    omConf[k] = config.pop(k)
            OptomechDevice.__init__(self, dm, config, name)
            DAQGeneric.__init__(self, dm, config, name)

        def getFilterDevice(self):
            # return parent filter device or None

acq4/devices/FilterWheel/FilterWheel.py
mgraupe: OK. got rid of all the import * 's

acq4/devices/FilterWheel/FilterWheel.py
mgraupe: OK.

acq4/devices/FilterWheel/FilterWheel.py
mgraupe: You are right.

acq4/devices/FilterWheel/FilterWheel.py
mgraupe: Right, I removed this signal emission.

acq4/devices/FilterWheel/FilterWheel.py
mgraupe: The method blocks until completion of the rotation. So the block you suggest is not necessary.

acq4/devices/FilterWheel/FilterWheel.py
mgraupe: right. I thought I had this implemented. I am using your line now.

acq4/drivers/ThorlabsFW102C/thorFW102cDriver.py
mgraupe: OK

acq4/drivers/ThorlabsFW102C/thorFW102cDriver.py
mgraupe: OK. implemented.

acq4/drivers/ThorlabsFW102C/thorFW102cDriver.py
mgraupe: OK

acq4/devices/FilterWheel/FilterWheel.py
mgraupe: Ok

acq4/modules/TaskRunner/analysisModules/Imaging/Imaging.py
mgraupe: OK

acq4/modules/Imager/Imager.py
mgraupe: 1. Corrected with a 'try' condtion 
2. Changed name to 'filterDevice' and signal to 'sigFilterChanged' to be more general.

acq4/modules/TaskRunner/analysisModules/Imaging/Imaging.py
mgraupe: Replaced by `self.params['scanner']`

acq4/modules/TaskRunner/analysisModules/Imaging/Imaging.py
mgraupe: Hi Luke,

thank you for all the input. I implemented everything except for your last
suggestion ... have to run now and will continue later.

Cheers,
Mike

On Sun, May 8, 2016 at 7:58 AM, Luke Campagnola <notifications@github.com>
wrote:

> In acq4/modules/TaskRunner/analysisModules/Imaging/Imaging.py
> <https://github.com/acq4/acq4/pull/45#discussion_r62427303>:
>
> >              ])
> >          self.ptree.setParameters(self.params, showTop=False)
> >          self.params.sigTreeStateChanged.connect(self.update)
> >          self.params.child('detectors').sigAddNew.connect(self.addDetectorClicked)
> >          self.params.child('decomb', 'auto').sigActivated.connect(self.autoDecomb)
> >
> >          self.man = getManager()
> > +        print self.man, dir(self.man.devices)
> > +        self.scannerDev = self.man.getDevice('Scanner')
> > +        # find first scope device that is parent of scanner
> > +        dev = self.scannerDev
> > +        while dev is not None and not isinstance(dev, Microscope):
> > +            dev = dev.parentDevice()
> > +        self.scopeDev = dev
> > +        self.filterwheel = self.man.getDevice('FilterWheel')
>
> Obviously this won't work on systems without a device with this name.. I
> think the correct approach here would be something like: 1) make the PMT an
> optomech device and set up the hierarchy like Stage -> Microscope ->
> FilterWheel -> PMT; 2) ask the PMT whether it has a filter selection in its
> parent devices. To start, you could do this pretty simply:
>
> class PhotoDetector(DAQGeneric, OptomechDevice):
>     def __init__(self, dm, config, name):
>         omConf = {}
>         for k in ['parentDevice', 'transform']:
>             if k in config:
>                 omConf[k] = config.pop(k)
>         OptomechDevice.__init__(self, dm, config, name)
>         DAQGeneric.__init__(self, dm, config, name)
>
>     def getFilterDevice(self):
>         # return parent filter device or None
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/acq4/acq4/pull/45/files/5a99fb27df27f5ded91a35f8bb9542993c512823#r62427303>
>



-- 
Michael Graupner (PhD)
Brain Physiology Lab, Université Paris Descartes
http://www.biomedicale.parisdescartes.fr/~mgraupe/
michael.graupner@parisdescartes.fr

mgraupe: I added a PMT device class and this class returns a parent FilterDevice. I used the same approach to implement the FilterDevice in the Imager module. 

What do you think of the implementation?

acq4/devices/PMT/PMT.py
```diff
@@ -0,0 +1,22 @@
...
+from acq4.devices.OptomechDevice import OptomechDevice
+from acq4.devices.DAQGeneric import DAQGeneric
+
+class PMT(DAQGeneric, OptomechDevice):
+    def __init__(self, dm, config, name):
+        self.omConf = {}
+        for k in ['parentDevice', 'transform']:
+            if k in config:
+                self.omConf[k] = config.pop(k)
+        OptomechDevice.__init__(self, dm, config, name)
```
campagnola: This one should use `omConf` instead of `config`.

acq4/devices/PMT/PMT.py
```diff
@@ -0,0 +1,22 @@
+# -*- coding: utf-8 -*-
+from acq4.devices.Device import *
+from acq4.devices.OptomechDevice import OptomechDevice
+from acq4.devices.DAQGeneric import DAQGeneric
+
+class PMT(DAQGeneric, OptomechDevice):
```
campagnola: Why `PMT` instead of `PhotoDetector`? I imagine exactly the same code could be used to cover photodiodes. I also have need for a device like this when calibrating lasers, so thermopile detectors might be included as well.

acq4/devices/PMT/PMT.py
mgraupe: `self.omConf` gives an error if placed instead of config:

`TypeError: string indices must be integers, not str`

acq4/devices/PMT/PMT.py
mgraupe: I didn't want to rename PMT everywhere else to PhotoDetector (e.g. for example as detector in the Imager module). Should I name the class PhotoDetector and rename PMT everywhere to PhotoDetector or is there another way?

campagnola: Let's go with `PhotoDetector` for the new class name and leave everything else as it is.

mgraupe: OK. 

And which relation of the current `PMT` device to `PhotoDetector` do you propose? I think this is the part which is not clear to me.

campagnola: I am only suggesting changing the class name on line 6 to `PhotoDetector`, and renaming the file/directory accordingly. There should be no need to change anything else in Imager or elsewhere in ACQ4, and you can still name your device "PMT" in the configuration file:

    PMT:
        driver: "PhotoDetector"

Maybe I still don't understand the question?

PR #46, opened 2016-08-26
Image analysis

campagnola:
@pbmanis:
I am making some changes to the image controls in the mosaic editor. Currently we have a list of filters (max w/ gaussian, tv denoise, etc.). Rather than continue adding to this list, I have replaced them with a flowchart that should let us construct and tune any number of new filter types as we need them. The downside is that what was previously a single click is now more like 2-3 clicks to load the flowchart you want for each image. I am open to feedback on this.

New flowchart nodes added to replace old filters: Data.max, .min, .mean, .std, .index, .slice

I am also converting the tv_denoise code you added into a flowchart node. However, I am curious about its provenance—it seems to be present in some forks of scikit-image, but not in the mainline. Can you remember where this code came from originally? Can we maybe import it rather than copying the code?


pbmanis:
I use the buttons frequently, as do others in the lab. When you are working with 2p stacks, it is very very convenient to get a median, mean, max or lightly filtered version that is the max to help with alignment against other sets of stacks, and checking the quality of the imaging on-line during experiments, and generating figures. Having to go through extra steps to do this for every set of stacks that one might collect (say, for reconstructing a cell from a set of 63x image stacks where you might have over a dozen stacks to work with) and having to deal with loading and debugging the flowchart to do it would be, in my mind, a big step backwards, and would make the operations more fragile (flowcharts don't always seem to be available on all systems, they get modified outside of the main repo, which creates a mess) as well as slower. I would really like to retain the base set of buttons. If you can do that while adding the ability to access a fc component, I think that would be optimal. I realize that handling the stacks was not where the mosaic editor started, but it's ability to do so in the context of spatial alignment is very nice.

WRT to tv_denoise, no I don't remember where I found it. However, it is not very useful and is slow so it could be pulled. An algorithm of the same name (it might even be the same or a precursor in any case) exists in the scikits image library.  Another algorithm that would be useful to put into the editor would be the Richardson-Lucy deconvolution method - I'm not sure you want to delegate that to a flowchart however, as effective and vetted code already exists.

campagnola:
Ok, I will think about ways to keep the streamlined workflow available. 

One option is to keep the buttons and have them directly modify the flowchart. This would avoid the issue of having the flowchart files only exist outside the repository, and add the benefit of being able to tweak the filter parameters if needed.

campagnola:
@pbmanis, I have added the buttons back in; they now load pre-configured flowcharts to implement their analysis. I have also added a button that lets you save the entire mosaic to file, which probably should have been done long ago. It's been pretty stable for me so I will probably merge it soon. Try it out and let me know what issues come up (either before or after the merge).

PR #50, opened 2016-12-08
Timecourse analyzer

meganbkratz:
This is a new analysis module for analyzing electrophysiology data repeated over time. (ie, multiple repetitions of the same protocol). It uses a flowchart to analyze the data, so is very flexible across different types of experiments. It can also save whatever data comes out of the flowchart into an acq4 sqlite database.


campagnola:
cooooooool!  I wish to try this.

campagnola:
Yaay!

PR #51, opened 2016-12-13
FileLoader Improvement

meganbkratz:
Adds a clear button to the file loader. Clicking the clear button calls host.clearFilesRequested; for backwards compatibility the clear button is disabled if host doesn't have a clearFilesRequested method.

Also moves the LoadFile button to above the splitter so that the user can hide the loadedFiles list and notes field while still having access to the LoadFile button.


campagnola:
Hooray!

PR #52, opened 2016-12-20
Camera tracking

campagnola:
Fixes camera module such that it correctly tracks stage movement when multiple cameras are present.

### DIFF
--- a/acq4/devices/Camera/CameraInterface.py
+++ b/acq4/devices/Camera/CameraInterface.py
@@ -33,11 +33,8 @@ class CameraInterface(CameraModuleInterface):
     def __init__(self, camera, module):
         CameraModuleInterface.__init__(self, camera, module)
 
-        self.module = module
-        self.view = module.getView()
         self.hasQuit = False
         self.boundaryItems = {}
-        self._trackView = False
 
         ## setup UI
         self.ui = CameraInterfaceTemplate()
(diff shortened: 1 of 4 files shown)


mgraupe:
I gave the camera-tracking branch a first test. Here are a couple of observations. 

In the Camera Module : 
- The white and the cyan ROI of one of the cameras does not move together, while they remain overlaying for the other camera. 
- When switching both cameras off, i.e., stopping "Aquire Video", the field of view jumps to a strange place which does not contain any image or ROI. It looks like as if the field of view jumps back to the original location. I think that it should remain at the new position. 

In the Imager Module : 
- The Multiphoton Imager does not load correctly (many panels and labels missing in the interface) and the following error occurs : 
===== 2016.12.22 12:12:32 =====
Traceback (most recent call last):
  File "acq4\modules\Manager\Manager.py", line 91, in loadModule
    self.manager.loadDefinedModule(mod)
  File "acq4\Manager.py", line 579, in loadDefinedModule
    mod = self.loadModule(mod, mName, config, forceReload=forceReload)
  File "acq4\Manager.py", line 512, in loadModule
    mod = modclass(self, name, config)
  File "acq4\modules\Imager\Imager.py", line 360, in __init__
    self.cameraModule.window().addInterface(self.name, self.camModInterface)
  File "acq4\modules\Camera\CameraWindow.py", line 154, in addInterface
    self.updateViewTracking()
  File "acq4\modules\Camera\CameraWindow.py", line 194, in updateViewTracking
    if iface.isRunning():
AttributeError: 'ImagerCamModInterface' object has no attribute 'isRunning'


Cheers,
Michael

PR #55, opened 2017-03-10
PyQtGraph missing function name fix

meganbkratz:
added missing import statement. 

getPreviousVersion is called in disconnect, but was not defined in the file or imported, so I added a statement to import it from reload.py


campagnola:
Ha. I just fixed this somewhere else.. now where'd I fix that?

campagnola:
This was fixed upstream: https://github.com/acq4/acq4/commit/89eac27c6124e1f27dacfd429541cc2577c2a97d

PR #56, opened 2017-03-10
Retiga R1 camera

meganbkratz:
added default settings for a new camera

### DIFF
--- a/acq4/drivers/pvcam/pvcam.py
+++ b/acq4/drivers/pvcam/pvcam.py
@@ -60,9 +60,16 @@ cameraDefaults = {
         ('binningX', 1, [1,2,3,4,8,12,16,24,32,48,64,128,256,512]),
         ('binningY', 1, [1,2,3,4,8,12,16,24,32,48,64,128,256,512]),
     ],
+
+    'RetigaR1': [
+        ('READOUT_PORT', 0),  ## Fastest port for R1
+        ('SPDTAB_INDEX', 2),  ## Fastest option for Q57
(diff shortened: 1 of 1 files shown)


acq4/drivers/pvcam/pvcam.py
```diff
@@ -60,9 +60,16 @@
         ('binningX', 1, [1,2,3,4,8,12,16,24,32,48,64,128,256,512]),
         ('binningY', 1, [1,2,3,4,8,12,16,24,32,48,64,128,256,512]),
     ],
+
+    'RetigaR1': [
+        ('READOUT_PORT', 0),  ## Fastest port for R1
+        ('SPDTAB_INDEX', 2),  ## Fastest option for Q57
```
campagnola: Is this the correct index for R1? (is the comment just left-over?)

meganbkratz: Oh, good question! I'll need to check with Paul.

pbmanis: I've poked around a bit - the R1 is pretty standard, and does have 2 ports (14 and 16 bit), and 3 speeds (the cameras come with a certfication for the readout noise for all 6 combinations of settings). Obviously the comment should no longer say Q57. I haven't found any documentation that precisely says how the port values are mapped to the speed or the readout, however.

meganbkratz: These settings have been in use for the past 4 months or so without problems. I propose that we keep the settings since they're working but remove both comments since we don't know that they are true.

PR #58, opened 2017-03-28
analysis fixes

meganbkratz:
Fixes for loading user configurations, and opening Photostim module (fixes for Spinbox.setOpts change)

### DIFF
--- a/acq4/Manager.py
+++ b/acq4/Manager.py
@@ -415,7 +415,7 @@ class Manager(QtCore.QObject):
             if name not in self.config['configurations']:
                 raise Exception("Could not find configuration named '%s'" % name)
             cfg = self.config['configurations'].get(name, )
-        self.configure(cfg)
+        self.configure({'storageDir':cfg})
 
     #def __del__(self):
         #self.quit()
--- a/acq4/Manager.py
+++ b/acq4/Manager.py
@@ -415,7 +415,7 @@ class Manager(QtCore.QObject):
             if name not in self.config['configurations']:
                 raise Exception("Could not find configuration named '%s'" % name)
             cfg = self.config['configurations'].get(name, )
-        self.configure({'storageDir':cfg})
+        self.configure(cfg)
 
     #def __del__(self):
         #self.quit()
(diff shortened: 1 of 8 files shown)


acq4/Manager.py
```diff
@@ -415,7 +415,7 @@ def loadDefinedConfig(self, name):
             if name not in self.config['configurations']:
                 raise Exception("Could not find configuration named '%s'" % name)
             cfg = self.config['configurations'].get(name, )
-        self.configure(cfg)
+        self.configure({'storageDir':cfg})
```
campagnola: I din't think this is right--shouldn't `cfg` already be a dict with `storageDir` inside?

acq4/pyqtgraph/widgets/SpinBox.py
```diff
@@ -143,7 +143,7 @@ def setOpts(self, **opts):
         """
         #print opts
         for k in opts:
-            if k == 'bounds':
+            if k == 'bounds' or k == 'range':
```
campagnola: Why? I don't think `range` was a valid option here in any version of pyqtgraph.

acq4/Manager.py
meganbkratz: You're right, I had my config file set up incorrectly. I changed the example config file to make it more clear how to set things up.

acq4/pyqtgraph/widgets/SpinBox.py
meganbkratz: I think you're right, but there are a lot of places (like in all the flowchart nodes) where we expect 'range' to be a valid option. These all probably failed silently in the past. Adding 'range' as an option to SpinBox seemed like a reasonable thing. Alternatively, we could change all the places we pass in range to use bounds instead. Thoughts?

campagnola: Let's just fix range->bounds wherever we see it. I'd prefer to keep the SpinBox API clean.

meganbkratz:
Hey Luke, 
Could you revisit merging this in? Thanks!
Megan

campagnola:
Thanks!

PR #59, opened 2017-03-29
small patch for cameras which don't support the setRoi function

mgraupe:
The ThorlabsUSBCamera implementation in MicroManager has not implemented the setRoi feature of the camera, for example. This small patch allows to use it in ACQ4 anyways.

### DIFF
--- a/acq4/devices/MicroManagerCamera/mmcamera.py
+++ b/acq4/devices/MicroManagerCamera/mmcamera.py
@@ -246,12 +246,21 @@ class MicroManagerCamera(Camera):
 
     def setParam(self, param, value, autoCorrect=True, autoRestart=True):
         return self.setParams({param: value}, autoCorrect=autoCorrect, autoRestart=autoRestart)
-
+    
+    def setRoi(self,val):
+        try:
+            self.mmc.setROI(*val)
+        except:
+            self.cameraSupportsROIs = False
+        else:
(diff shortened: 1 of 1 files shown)


acq4/devices/MicroManagerCamera/mmcamera.py
```diff
@@ -246,12 +246,21 @@ def setParams(self, params, autoRestart=True, autoCorrect=True):
 
     def setParam(self, param, value, autoCorrect=True, autoRestart=True):
         return self.setParams({param: value}, autoCorrect=autoCorrect, autoRestart=autoRestart)
-
+    
+    def setRoi(self,val):
+        try:
+            self.mmc.setROI(*val)
+        except:
```
campagnola: I think this will lead to frustrated users wondering why nothing happens when they try to set their ROI (especially if the camera does support that feature, but setting the ROI failed for another reason). Perhaps MicroManager provides a mechanism to explicitly ask whether the camera supports ROIs?

If we can find a safe way to determine ROI support, then the "correct" way to disable this should be through the `MicroManagerCamera.listParams()` method (here: https://github.com/acq4/acq4/blob/develop/acq4/devices/MicroManagerCamera/mmcamera.py#L197  ..either it should list the parameter as read-only, or it should limit its range of allowable values such that it cannot be changed).

PR #67, opened 2017-11-01
TDT

meganbkratz:
This PR is mostly for TDT Device work, however other miscellaneous things got mixed in along the way, and I'm not sure how best to deal with that.


acq4/devices/Scanner/scan_program/rect.py
```diff
@@ -21,7 +21,8 @@ class RectScanComponent(ScanProgramComponent):
     def __init__(self, scanProgram=None):
         ScanProgramComponent.__init__(self, scanProgram)
         self.ctrl = RectScanControl(self)
-        
+        print('recscan INIT')
```
campagnola: remove

meganbkratz: Done, sorry.

PR #71, opened 2017-12-29
Change to TimeoutError

arrrobase:
sci = self.send('scientifica', timeout=0.2) (line 132) will lead to a TimeoutError, not RunetimeError. If want to try again need to catch TimeoutError.


acq4/drivers/Scientifica/scientifica.py
```diff
@@ -130,7 +130,7 @@ def __init__(self, port=None, name=None, baudrate=None, ctrl_version=2):
                 SerialDevice.__init__(self, port=self.port, baudrate=baudrate)
                 try:
                     sci = self.send('scientifica', timeout=0.2)
-                except RuntimeError:
+                except TimeoutError:
```
campagnola: In some circumstances we might get a RuntimeError as well, so it should be

    except (RuntimeError, TimeoutError):

PR #74, opened 2019-08-13
Change GS algorithm

migtown:
Change GS algorithm; pull request test


campagnola:
Hey @migtown, have a look at the "files changes" tab above -- I don't think these are the changes that you intended to make a PR for..

PR #75, opened 2020-03-04
add micromanager device unloading into Manager.quit() method

berl:
This adds back the micromanager device unloading into the `Manager` via altered `utils.micromanager` 

Could use some testing on other systems and configurations...

### DIFF
--- a/acq4/Manager.py
+++ b/acq4/Manager.py
@@ -32,6 +32,11 @@ from .util.HelpfulException import HelpfulException
 from . import __version__
 from . import devices, modules
 
+
+from .util.micromanager import unloadMMCore
+
+
+
 ### All other modules can use this function to get the manager instance
 def getManager():
     if Manager.single is None:
@@ -837,7 +842,14 @@ class Manager(Qt.QObject):
                         printExc("Error while requesting device '%s' quit." % d)
                     #print "  done."
                     dlg.setValue(lm+ld-len(self.devices))
-                    
+
+                print(globals().keys())
+                try:
+                    unloadMMCore()
+                except:
+                    print("failed to unloadMMCore")
+                print("Goodbye")
+
                     
                 print("Closing windows..")
                 Qt.QApplication.instance().closeAllWindows()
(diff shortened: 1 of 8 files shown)


acq4/Manager.py
```diff
@@ -837,7 +842,14 @@ def quit(self):
...
                     #print "  done."
                     dlg.setValue(lm+ld-len(self.devices))
-                    
+
+                print(globals().keys())
+                try:
+                    unloadMMCore()
+                except:
+                    print("failed to unloadMMCore")
+                print("Goodbye")
```
campagnola: Maybe do this with `atexit` rather than by modifying Manager?

acq4/util/micromanager.py
```diff
@@ -2,16 +2,47 @@
...
+DEFAULT_MM_PATH = "C:\\Program Files\\Micro-Manager-1.4"
+
+def getMMConfig():
+    """
+        get the micromanager path and desired micromanager config file from acq4 config 
+    """    
+    for path in CONFIGPATH:
+        cf = os.path.join(path, 'default.cfg')
+        if os.path.isfile(cf):
+            cfg = configfile.readConfigFile(cf)
```
campagnola: Rather than re-reading the config file, maybe the Manager should have a method for accessing this, like `getConfig()` ?

acq4/util/micromanager.py
```diff
@@ -31,7 +62,38 @@ def getMMCorePy(path=None):
...
+            try:
+                _mmc.loadSystemConfiguration(mm_config_file)
+                _mmc.getVersionInfo()
+            except:
+                print(_mmc.getDeviceAdapterSearchPaths())
+                raise ValueError("micromanager loadSystemConfiguration failed!")
+        else:
+            print("no micromanager configuration defined or loaded")
+    else:
+        print("micromanager configuration already loaded")
```
campagnola: Seems like this should only be called once when CMMCore is instantiated, rather than every time getMMCorePy is called.

acq4/util/micromanager.py
```diff
@@ -31,7 +62,38 @@ def getMMCorePy(path=None):
...
+
     return _mmc
+
+
+def unloadMMCore():
+    print("attempting to unload Micro Manager Devices ... ")
+    print(globals().keys())
+    if "_mmc" in globals().keys():
```
campagnola: `_mmc` is defined as None at the top of this file; should be no need to check.

PR #80, opened 2020-06-05
µManager binning-roi scaling

outofculture:
This behavior first shows up in MMCore version 7.0.2 (distributed as MicroManager 1.4.21). For all versions after that (including 2.0), use binned pixel units in ROI operations.

### DIFF
--- a/acq4/devices/MicroManagerCamera/mmcamera.py
+++ b/acq4/devices/MicroManagerCamera/mmcamera.py
@@ -258,28 +258,27 @@ class MicroManagerCamera(Camera):
             p('start')
         
         needRestart = False
-        return (newVals, needRestart)
+        return newVals, needRestart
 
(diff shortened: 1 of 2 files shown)


acq4/devices/MicroManagerCamera/mmcamera.py
```diff
@@ -206,6 +207,38 @@ def _readAllParams(self):
 
             self._allParams = params
 
+    def getROI(self):
+        cam_region = self.mmc.getROI(self.camName)
```
campagnola, line 211: I know it hurts, but we should prolly `camRegion` for consistency.

outofculture: Oh man; I didn't even notice I did that. Yep.

PR #81, opened 2020-07-07
Sensapex pressure controller v0.915

outofculture:
I couldn't come up with a way to test that I am converting kPa to Pa correctly, so double-check my logic there.

Note: the sdk offers two `get_pressure` functions, one for the setting, and one for an actual measurement. I picked the "setting" one. We only ever check the pressure when we init the `SensapexPressureControl`.

### DIFF
--- a/acq4/drivers/sensapex/sensapex.py
+++ b/acq4/drivers/sensapex/sensapex.py
@@ -600,13 +600,19 @@ class SensapexDevice(object):
 
     def set_pressure(self, channel, value):
         """
-        @param value: pressure in kPa
+        Parameters
+        ----------
+        value : float
+            pressure in kPa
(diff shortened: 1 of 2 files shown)


acq4/drivers/sensapex/sensapex.py
```diff
@@ -597,9 +599,15 @@ def _change_callback(self, devid, new_pos, old_pos):
             self.callback(self, new_pos, old_pos)
 
     def set_pressure(self, channel, value):
-        return self.ump.set_pressure(self.devid, int(channel), int (value))
+        """
+        @param value: pressure in kPa
```
campagnola, line 603: We've been using numpy-style docstrings here.

PR #82, opened 2020-07-20
Zeiss shutter control

outofculture:
This also adds error reporting for zeiss event handlers.


acq4/devices/LightSource/LightSource.py
```diff
@@ -32,7 +32,7 @@ def __init__(self, dm, config, name):
         self._lock = Mutex.Mutex()
 
     def deviceInterface(self, win):
-        return LightSourceGui(self)
+        return LightSourceDevGui(self)
 
     def taskInterface(self, taskRunner):
         return LightSourceTaskGui(self, taskRunner)
```
campagnola, line 38: just return None until LightSourceTaskGui gets filled in.

PR #86, opened 2020-08-11
SignalProxy.block needs to be renamed

outofculture:
The rename PR in pyqtgraph - https://github.com/pyqtgraph/pyqtgraph/pull/1344 - will rename this out from under us


outofculture:
Oops. This was due to a stale pyqtgraph branch in use on my test machine. Disregard.

PR #87, opened 2020-08-17
sensapex drivers are gone. long live sensapex!

outofculture:
Moved sensapex sdk wrapper into its own package.


campagnola:
Looks good. One request: keep acq4.drivers.sensapex and just raise an exception letting people know where the code has gone. It might also help to something like

    try:
        from sensapex import UMP
    except ImportError:
        raise Exception("Could not import sensapex driver package (acq4 now requires sensapex-py to be installed separately)")

outofculture:
@campagnola how's this?

PR #95, opened 2020-10-06
Put image processing in a thread

outofculture:
Contrast and BG controllers are made thread-safe. MMC cameras might need read-only protection.


campagnola:
Seems like threading is not going to help much in this location..  we can close this for now at least?

PR #96, opened 2020-10-07
Autopatch foundlings

outofculture:
I was looking for authorship information on the voltage-imaging/autopatch code, and I found this branch with some changes which might be useful? Only a few lines are real code changes, but they should maybe get looked at before they fall away to the sands of time.


campagnola:
I seem to remember we were working on improving pipette detection and abandoned this branch because it wasn't improving anything. ..in any case, it can't be merged without some way to test it so I suggest just closing.

PR #98, opened 2020-10-16
c library fixes

outofculture:
3 unrelated updates are included here.

clibrary fixes: the changes are mostly the same as what @campagnola figured out with me the other day. The new cache format doesn't deserialize with tuples (maybe it should include an explicit type somewhere, eventually). Multiclamp code needs bytes instead of strings.

`--exit-on-error` flag to make reading errors easier.

Refactor of how frames-per-second limits are written in the code.

### DIFF
--- a/acq4/Manager.py
+++ b/acq4/Manager.py
@@ -82,6 +82,7 @@ class Manager(Qt.QObject):
         self.config = OrderedDict()
         self.currentDir = None
         self.baseDir = None
+        self.exitOnError = False
         self.gui = None
         self.shortcuts = []
         self.disableDevs = []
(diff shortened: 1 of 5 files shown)


acq4/util/imaging/frame_display.py
```diff
@@ -8,6 +8,10 @@
 from .contrast_ctrl import ContrastCtrl
 
 
+MAX_FRAMES_PER_SECOND = 30
+S_PER_FRAME = 1.0 / MAX_FRAMES_PER_SECOND
+MS_PER_FRAME = int(S_PER_FRAME * 1000)
```
outofculture, line 13: I don't love that these two vars have similar names but different data types.

acq4/Manager.py
```diff
@@ -103,13 +104,14 @@ def __init__(self, configFile=None, argv=None):
...
-                        argv, 'c:a:m:b:s:d:nD',
+                        argv, 'c:a:x:m:b:s:d:nD',
                         ['config=', 'config-name=', 'module=', 'base-dir=', 'storage-dir=',
-                         'disable=', 'no-manager', 'disable-all'])
+                         'disable=', 'no-manager', 'disable-all', 'exit-on-error'])
                 except getopt.GetoptError as err:
                     print(str(err))
                     print("""
     Valid options are:
+        -x --exit-on-error Whether to exit immidiately on the first exception
```
campagnola, line 114: **during initial manager configuration

acq4/util/imaging/frame_display.py
campagnola: Why global variables rather than instance variables?

PR #107, opened 2020-12-18
CoolLED light source support

outofculture:
Also fixes a bug in the base LightSource where all the sources only ever use the name of the last source added.


outofculture:
Untested until I have access to working hardware again...

Issue #114, opened 2021-05-12
TaskRunner Dock


pbmanis:
acq4 Branch: Python3
Pyqtgraph: 0.12.1
The TaskRunner is inconsistent in updating the devices displayed in the lower dock region after loading a new protocol. Either all devices display (e.g., DAQ, MultiClamp1), or the dock area is completely blank. Resizing the dock using the divider between the upper and lower regions does force the display to update.

PR #115, opened 2021-05-12
Python3 manislab: 

pbmanis:
A set of small cumulative fixes to get MultiClamp 700B, PVCam cameras, and NIDAQ support working under Python3. Also minor changes to TaskRunner/SequenceRunner for Python3.

- commit #2c1f556 handles arguments/order for nidaq/pydaqmx. Solves  Issue (closed now) raised in December.
- commit #3d8616 fixes ring buffer size calculation in PVCam.py device, and pvcam.py driver. It also corrects call to LIB rather than pvcam.call (longstanding issue often manually corrected).
- commit #26127ac changes MultiClamp TaskGUI quit call to be PyQt5 compatible, and to be consistent with other calls.
- commit 613b5ab util/SequenceRunner.py : forces generation of the range values, rather than returning the generator (python 2 vs python 3)
- commit #d24d88b PatchWindow: update for numpy.min/max: requires array-like (so cast to tuple)
- commit #3a0da18 Removes a lagging print statement in drivers/Multiclamp/multiclamp.py (may not exist upstream)
- commit #c7c40a2  Fixes a signal name in MultiClamp TaskGUI (related to #26127ac)

Tested with MC700B, Retiga R1 in a basic configuration. Issue #114 regarding TaskRunner dock behavior is not addressed by these changes.

### DIFF
--- a/acq4/analysis/dataModels/PatchEPhys/PatchEPhys.py
+++ b/acq4/analysis/dataModels/PatchEPhys/PatchEPhys.py
@@ -753,7 +753,7 @@ class GetClamps():
             if len(sequence_values) > 0:
                 self.values.append(sequence_values[i])
             else:
-                self.values.append(cmd[len(cmd) / 2])
+                self.values.append(cmd[int(len(cmd) / 2)])
         if traces is None or len(traces) == 0:
             print("PatchEPhys/GetClamps: No data found in this run...")
             return None
--- a/acq4/analysis/modules/IVCurve/IVCurve.py
+++ b/acq4/analysis/modules/IVCurve/IVCurve.py
@@ -1063,7 +1063,7 @@ class IVCurve(AnalysisModule):
         dt = self.Clamps.sample_interval
         rgnindx = [int((rgnpk[1]-0.005)/dt), int((rgnpk[1])/dt)]
         rmps = self.ivbaseline
-        vmeans = np.mean(self.Clamps.traces[:, rgnindx[0]:rgnindx[1]], axis=1) - self.ivbaseline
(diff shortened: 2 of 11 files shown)


pbmanis:
Somehow the branch "Python3-manislab" includes 2 fixes to the Sutter MPC200 device and driver to enable communication in Python3. These were supposed to be part of another branch "Python3-manislab-1" that addresses additional issues with a variety of specific devices on "Rig3" (Luke knows what that means). I'll post those as a PR after I have exercised them more, so the two Sutter-related changes are optional to incorporate right now.

campagnola:
Many thanks, Paul! 
One question: do you remember what nidaq problems you ran into that required those fixes? I'd love to write some tests..

pbmanis:
Hi Luke,

Long delayed response:

The code would crash in nidaq.py because the format of the call to the pydaqmx library was not always correct (of the correct number of arguments). If I recall correctly, this happened after an update in nidaqmx, around V20. This fix took a few more steps when building the argument list for the function call to try to better handle the case of “assume 1 missing argument” (not assume…), and to handle the case of calls with a “reserved” argument.
This has been in intermittent use on two of the rigs with no problems for a couple of months now. The third rig is still running python2.7 for pandemic-related reasons, but I am going to have to force an update there when I get a few free days.

So tests should confirm that calls in nidaq.py handle the possible different signatures and generate a successful call to the function with the arguments.

Meanwhile, I have been (very low priority) working on re-incorporating the automated tiling and stack recording into the camera module (removed from the Imaging module from way back). Still have a few loose ends before that will be ready.

Hope all is well!

--Paul

On 7/27/21, 5:45 PM, "Luke Campagnola" ***@***.***> wrote:


Many thanks, Paul!
One question: do you remember what nidaq problems you ran into that required those fixes? I'd love to write some tests..

PR #123, opened 2021-10-08
More manipulator/stage distinction

outofculture:
This belonged in #122, but sometimes commits get lost.


outofculture:
Ack, no, #122 was merged into `develop` instead of `main`. That's the problem.

PR #126, opened 2021-11-11
Fix bugs found when attempting to record clamp signals

outofculture:
Fix a divide-by-zero. Fix binary file handling. Cleanup a few things along the way.

### DIFF
--- a/acq4/devices/PatchPipette/testpulse.py
+++ b/acq4/devices/PatchPipette/testpulse.py
@@ -349,15 +349,16 @@ class TestPulse(object):
                 Vc = params['amplitude']
                 Rs = analysis['peakResistance']
                 Rm = Rin - Rs
-                Rs_denom = (Rm**2 * abs(Vc))
-                if Rs_denom != 0.0:
-                    analysis['capacitance'] = (Rin**2 * Q) / Rs_denom
+                Cm_denom = (Rm**2 * abs(Vc))
+                if Cm_denom != 0.0:
+                    analysis['capacitance'] = (Rin**2 * Q) / Cm_denom
                 else:
                     analysis['capacitance'] = 0
-            elif analysis['steadyStateResistance'] > 0:
-                analysis['capacitance'] = tau / analysis['steadyStateResistance']
-            else:
-                analysis['capacitance'] = np.nan
+            else:  # IC mode
+                if analysis['steadyStateResistance'] > 0:
+                    analysis['capacitance'] = tau / analysis['steadyStateResistance']
+                else:
+                    analysis['capacitance'] = np.nan
 
             # # detect bad fits
             # noise = (pulseData - scipy.ndimage.gaussian_filter(pulseData, 3)).std()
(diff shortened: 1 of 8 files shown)


acq4/devices/PatchPipette/testpulse.py
```diff
@@ -353,22 +347,17 @@ def analysis(self):
                 Q = (pulseData - yoffset).sum() * dt
                 Rin = analysis['steadyStateResistance']
                 Vc = params['amplitude']
-                Rs_denom = (Q * Rin + tau * Vc)
+                Rs = analysis['peakResistance']
+                Rm = Rin - Rs
+                Rs_denom = (Rm**2 * abs(Vc))
```
campagnola, line 352: Let's call this Cm_denom since it's used for calculating membrane capacitance

acq4/devices/PatchPipette/testpulse.py
```diff
@@ -353,22 +347,17 @@ def analysis(self):
...
-                analysis['capacitance'] = Cm
+                    analysis['capacitance'] = 0
+            elif analysis['steadyStateResistance'] > 0:
+                analysis['capacitance'] = tau / analysis['steadyStateResistance']
             else:
-                if analysis['steadyStateResistance'] > 0:
-                    analysis['capacitance'] = tau / analysis['steadyStateResistance']
-                else:
-                    analysis['capacitance'] = np.nan
+                analysis['capacitance'] = np.nan
```
campagnola, line 360: The original structure had the if-block for VC and the else-block for IC; I think it's more readable as it was?

acq4/devices/Pipette/tracker.py
```diff
@@ -539,14 +541,14 @@ def mapErrors(
         print("Manipulator missed target %d times" % misses)
 
         filename = self.dev.configFileName("error_map.np")
-        np.save(open(filename, "wb"), self.errorMap)
+        np.save(filename, self.errorMap)
```
campagnola, line 544: Are you sure this is the same as `open(..., 'wb')` ?

acq4/devices/Pipette/tracker.py
```diff
@@ -539,14 +541,14 @@ def mapErrors(
...
 
         return self.errorMap
 
     def showErrorAnalysis(self):
         if not hasattr(self, "errorMap"):
             filename = self.dev.configFileName("error_map.np")
-            self.errorMap = np.load(open(filename, "rb"))[np.newaxis][0]
+            self.errorMap = np.load(filename)[np.newaxis][0]
```
campagnola, line 551: likewise here

acq4/filetypes/Analyze75.py
```diff
@@ -231,11 +220,11 @@ def parseNii(headerFH, imgFile):
             data = headerFH.read(size)
         elif m.magic == 'nii\0':               ## data is in a separate .img file
             imgFile = os.path.splitext(imgFile)[0] + '.img'
-            fh = open(imgFile, 'rb')
-            fh.seek(m.vox_offset)
-            data = fh.read(size)
+            with open(imgFile, 'r', encoding="utf-8") as fh:
```
campagnola, line 223: Sure about this change? We don't have a test for it..

acq4/devices/Pipette/tracker.py
outofculture: That's what in the numpy source in at least 1.20.2, yeah.

acq4/filetypes/Analyze75.py
outofculture: `np.fromstring` wants strings, not bytes.

campagnola: But `data` is likely binary, not unicode. That means opening as a unicode string will likely just fail. I think the correct fix is to use `np.frombuffer` anywhere we were previously using `fromstring`.

PR #130, opened 2022-02-08
set restype on ctypes function objects

outofculture:
also do some maintenance.

### DIFF
--- a/acq4/util/clibrary/CParser.py
+++ b/acq4/util/clibrary/CParser.py
@@ -398,6 +398,7 @@ class CParser:
         for i in range(len(lines)):
             line = lines[i]
             newLine = ''
+            macroName = None
             m = directive.match(line)
             if m is None:  # regular code line
                 if ifTrue[-1]:  # only include if we are inside the correct section of an IF block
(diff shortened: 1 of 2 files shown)


acq4/util/clibrary/CParser.py
```diff
@@ -482,7 +483,7 @@ def pa(t):
                 elif d == 'pragma':  ## Check for changes in structure packing
                     if not ifTrue[-1]:
                         continue
-                    m = re.match(r'\s+pack\s*\(([^\)]+)\)', rest)
+                    m = re.match(r'\s+pack\s*\(([^)]+)\)', rest)
```
campagnola, line 486: why?

outofculture: Have you seen the rent on a backslash? Those things are easily 20% more expensive than other characters.

(Really, this was part of the "hey pycharm, are there any egregious errors in this file?" check, and it included this unnecessary escape in said category and I said "sure, buddy; if you say so." and didn't give it any more thought.)

campagnola: Oh oh I see, the escape is unnecessary inside `[]`

PR #133, opened 2022-02-14
position needs to be set after adding the ellipse to the window

outofculture:
I don't know what change precipitated this bug; some Qt thing or another, maybe? I don't have an environment where this change doesn't work, but it might need to be tested on an older install.


campagnola:
Oh, that's bad news.. that potenatially suggests a LOT of new bugs.

PR #142, opened 2023-06-03
Fix remaining deprecated time calls.

marsiwiec:
There were two remaining prime calls waiting to be fixed in order for my code to work after the recent changes.


outofculture:
Thanks! These are fixed now in main.

PR #175, opened 2023-09-29
Fix Reseal to actually try to seal

### LINKED ISSUE #157: Inaccurate pipette state after reseal completed
State field in Multipatch changes to Whole Cell instead of "Out" or In Bath to accurately reflect pipette above tissue surface.

outofculture:
Closes #157

### DIFF
--- a/acq4/devices/PatchPipette/states.py
+++ b/acq4/devices/PatchPipette/states.py
@@ -1136,7 +1136,7 @@ class PatchPipetteResealState(PatchPipetteState):
         'retractionSpeed': {'type': 'float', 'value': 0.3e-6, 'suffix': 'm/s'},
         'resealTimeout': {'type': 'float', 'value': 10 * 60, 'suffix': 's'},
         'numTestPulseAverage': {'type': 'int', 'value': 3},
-        'fallbackState': {'type': 'str', 'value': 'whole cell'},
+        'fallbackState': {'type': 'str', 'value': 'bath'},
         'maxPressure': {'type': 'float', 'value': -4e3, 'suffix': 'Pa'},
         'pressureChangeRate': {'type': 'float', 'value': -0.5e-3 / 60, 'suffix': 'Pa/s'},
     }
--- a/acq4/devices/PatchPipette/states.py
+++ b/acq4/devices/PatchPipette/states.py
@@ -1110,12 +1110,12 @@ class PatchPipetteResealState(PatchPipetteState):
     ----------
     initialPressure : float
         Initial pressure (Pa) to apply (default is -0.5 kPa)
-    maximumPressure : float
+    maxPressure : float
         Maximum pressure (Pa) to apply (default is -4 kPa)
     pressureChangeRate : float
         Rate at which pressure should change during reseal (default is -0.5 kPa / min)
     retractionSpeed : float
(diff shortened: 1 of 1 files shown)


acq4/devices/PatchPipette/states.py
```diff
@@ -1065,9 +1065,10 @@ class PatchPipetteResealState(PatchPipetteState):
         'retractionSpeed': 0.3e-6,
         'resealTimeout': 10 * 60,
         'numTestPulseAverage': 3,
-        'fallbackState': 'whole cell',
+        'fallbackState': 'bath',
```
campagnola, line 1068: 'whole cell' is the correct fallback state -- we are starting from whole cell; if anything unexpected happens while resealing, then we should return to whole cell. Falling back to bath means that a bug in the reseal procedure causes the cell to be instantly killed.

acq4/devices/PatchPipette/states.py
```diff
@@ -1120,7 +1122,7 @@ def run(self):
                 patchrec['resealInitialResistance'] = initialResistance
 
             # check progress on resistance
-            if len(recentTestPulses) > config['numTestPulseAverage']:
+            if len(recentTestPulses) == config['numTestPulseAverage']:
```
campagnola, line 1125: if len(recentTestPulses) == recentTestPulses.maxlen:

acq4/devices/PatchPipette/states.py
```diff
@@ -1265,7 +1267,7 @@ def run(self):
         dev.clean = True
         self.resetPosition()
         dev.newPatchAttempt()
-        return 'out'          
+        return config['fallbackState']
```
campagnola, line 1270: why?

acq4/devices/PatchPipette/states.py
outofculture: I was told that when reseal fails, the pipette is in the bath, no longer attached to the cell.

acq4/devices/PatchPipette/states.py
outofculture: fallbackState is 'out', and isn't used ever, so I had just imagined they were meant to go together.

outofculture: "isn't used ever"... in this class. the base class sets `self.nextState` to it, so it is still in use there.

acq4/devices/PatchPipette/states.py
```diff
@@ -1041,16 +1041,18 @@ class PatchPipetteResealState(PatchPipetteState):
...
         Rate at which pressure should change during reseal (default is -0.5 kPa / min)
     retractionSpeed : float
-        Speed in m/s to move pipette during retraction (default is 0.3 um / s)
+        Speed in m/s to move pipette during retraction (default is 0.3e-6 (0.3 µm / s))
     resealTimeout : float
         Seconds before reseal attempt exits
     numTestPulseAverage : int
         Number of test pulses to average when measuring resistance
+    slowDetectionThreshold : float
+        Threshold for slow change in pipette resistance (Ohm) to trigger cell detection (default 200 kOhm)
```
campagnola, line 1055: To trigger reseal detection?

outofculture: Maybe? I didn't actually grok the code, I just noticed it was used at all but not included in the defaults or docs.

PR #185, opened 2023-10-11
live-editable patch profiles

### LINKED ISSUE #162: Add UI for editing patchpipette state parameters
Need a way to modify state parameters in mid-experiment
- Reusable Qt elements (probably parametertree)
- Easy access from multipatch module
- Show eatch patch profile including copy-from field
- A value that is not present in config file should show up as "default" in parametertree
- Maybe allow writing config changes back to disk?

outofculture:
This adds a live editor for patch profiles.

Closes #162


acq4/modules/MultiPatch/patchProfileEditor.py
```diff
@@ -0,0 +1,118 @@
...
+            self.addChild(StateParameter(state, profile))
+
+    def reinitialize(self):
+        for child in self:
+            if isinstance(child, StateParameter):
+                child.reinitialize()
+
+    def applyDefaults(self, defaults):
+        for state in self:
+            if state.name() in defaults:
```
outofculture, line 80: @campagnola here is an example of the code I don't like. I expect to be able to just do `self[key] for key in defaults`, but that doesn't work.

PR #197, opened 2024-01-30
Fill in real reseal logic

### LINKED ISSUE #159: Reseal: pause / reverse when resistance fluctuates
Needed modification to the reseal procedure: 
- Monitor input resistance, get a baseline + variance 
  - Perhaps this requires just looking over the history for this cell? Perhaps we record for 20 seconds to collect this?
- While retracting pipette, react to significant changes in resistance:
  - Increase in resistance indicates that the membrane is stretching. Pause, advance pipette a few microns, and wait for resistance to stabilize before continuing retraction
  - Decrease in resistance indicates membrane is tearing. Pause, retract, and wait for resistance to recover. Potentially retract as far as the original start location.

outofculture:
closes #159


acq4/devices/PressureControl/device.py
```diff
@@ -31,17 +34,54 @@ def __init__(self, manager, config, name):
...
+        rate: Optional[float] = None,
+        _future: Optional[Future] = None,
+    ) -> None:
+        if target is None and maximum is None and minimum is None:
+            raise ValueError("Must specify at least one of target, maximum, or minimum")
+        if target is not None and (maximum is not None or minimum is not None):
+            raise ValueError("Cannot specify both target and maximum/minimum")
+        # TODO do we need to guarantee that the source gets set?
+
+        def value_is_out_of_bounds(val):
```
campagnola, line 53: Something like this? It's not dramatically shorter, but feels more straightforward to me.

```
if target is not None:
    minimum = target - target_tolerance
    maximum = target + target_tolerance

startPressure = endPressure = self.getPressure()
if minimum is not None:
    endPressure = max(minimum, endPressure)
if maximum is not None:
    endPressure = min(maximum, endPressure)

startTime = ptime.time()   # time.time() works differently on windows! 
endTime = (endPressure - startPressure) / rate
fracDone = 0
while fracDone < 1:
    fracDone = min((ptime.time() - startTime) / (endTime - startTime), 1)
    self.setPressure(startPressure + fracDone * (endPressure - startPressure))
    _future.sleep(self.regulatorSettlingTime) 
```

acq4/devices/PressureControl/device.py
```diff
@@ -31,16 +32,37 @@ def __init__(self, manager, config, name):
...
+            end_pressure = max(minimum, end_pressure)
+        if maximum is not None:
+            end_pressure = min(maximum, end_pressure)
+
+        start_time = ptime.time()
+        end_time = start_time + abs(end_pressure - start_pressure) / rate
+        frac_done = 0
+        while frac_done < 1:
+            frac_done = min((ptime.time() - start_time) / (end_time - start_time), 1)
+            print(frac_done)
```
campagnola, line 64: print!

acq4/devices/PressureControl/device.py
```diff
@@ -31,16 +32,37 @@ def __init__(self, manager, config, name):
...
-        self.setPressure(source)
-        while True:
-            pressure = self.getPressure()
-            if minimum is not None and pressure < minimum:
-                self.setPressure(source, pressure=minimum)
-            elif maximum is not None and pressure > maximum:
-                self.setPressure(source, pressure=maximum)
-            else:
-                break
+    def attainPressure(
```
campagnola, line 35: what about `rampPressure` ?

acq4/devices/PatchPipette/states.py
```diff
@@ -1130,97 +1161,257 @@ class PatchPipetteResealState(PatchPipetteState):
...
+
+        prev_dir = random_wiggle_direction()
+        for _ in range(self.config['nuzzleRepetitions']):
+            self.dev.pressureDevice.setPressure(source='regulator', pressure=self.config['nuzzleInitialPressure'])
+            self._pressureFuture = self.dev.pressureDevice.attainPressure(
+                target=self.config['nuzzlePressureLimit'], rate=self.config['nuzzlePressureChangeRate'])
+            start = ptime.time()
+            while ptime.time() - start < self.config['nuzzleDuration']:
+                while np.dot(direction := random_wiggle_direction(), prev_dir) > 0:
+                    pass  # ensure different direction from previous
```
campagnola, line 1227: Isn't this so unlikely as to be unnecessary?

acq4/devices/PatchPipette/states.py
```diff
@@ -1130,97 +1161,257 @@ class PatchPipetteResealState(PatchPipetteState):
...
+            """pick a random point on a circle perpendicular to the pipette axis"""
+            while np.linalg.norm(vec := np.cross(pipette_direction, np.random.uniform(-1, 1, size=3))) == 0:
+                pass  # prevent division by zero
+            return self.config['nuzzleLateralWiggleRadius'] * vec / np.linalg.norm(vec)
+
+        prev_dir = random_wiggle_direction()
+        for _ in range(self.config['nuzzleRepetitions']):
+            self.dev.pressureDevice.setPressure(source='regulator', pressure=self.config['nuzzleInitialPressure'])
+            self._pressureFuture = self.dev.pressureDevice.attainPressure(
+                target=self.config['nuzzlePressureLimit'], rate=self.config['nuzzlePressureChangeRate'])
```
campagnola, line 1223: If attainPressure had a duration argument (instead of or in addition to rate), then we could drop the nuzzlePressureChangeRate parameter

acq4/devices/PatchPipette/states.py
```diff
@@ -1130,97 +1161,257 @@ class PatchPipetteResealState(PatchPipetteState):
...
+    def unfreezeRollingResistanceThresholds(self):
+        self._updateRollingResistanceThresholds = True
 
     def run(self):
         config = self.config
         dev = self.dev
-        self.monitorTestPulse()
+        if config['extractNucleus'] is True:
+            self.sealToNucleus()
+        dev.pressureDevice.setPressure(config['initialPressureSource'], config['initialPressure'])
```
campagnola, line 1347: Why this? Shouldn't sealToNucleus leave us at the desired pressure? (and if it isn't called, then we are already at the initial pressure anyway)

acq4/devices/PatchPipette/states.py
```diff
@@ -1130,97 +1161,257 @@ class PatchPipetteResealState(PatchPipetteState):
...
+        self.checkStop()
+        self.setState("measuring baseline resistance")
+        self.startRollingResistanceThresholds()
```
campagnola, line 1350: Since this takes some time to populate, it makes sense to start it before nuzzling.

acq4/devices/PatchPipette/states.py
```diff
@@ -1130,97 +1161,257 @@ class PatchPipetteResealState(PatchPipetteState):
...
+    def analyzeRecentTestPulses(self):
+        analysis = [tp[1] for tp in self._rollingRecentTestPulses]
+        access_res = np.fromiter((a['peakResistance'] for a in analysis), dtype=float)
+        input_res = np.fromiter((a['steadyStateResistance'] - a['peakResistance'] for a in analysis), dtype=float)
+        self._recentInputResistanceMean = np.mean(input_res)
+        self._recentInputResistanceVariance = np.var(input_res)
+        self._recentAccessResistanceMean = np.mean(access_res)
+        self._recentAccessResistanceVariance = np.var(access_res)
+
+    def freezeRollingResistanceThresholds(self):
```
campagnola, line 1335: Why are these needed?

acq4/devices/PatchPipette/states.py
outofculture: no, `dot` will be `> 0` for all vectors in the same semicircle as each other, so this is making sure that the two directions are significantly different, not just "not identical".

acq4/devices/PatchPipette/states.py
outofculture: That would be great, yeah. Will those baseline measurements be valid, if they're taken while the pipette is busy extracting the nucleus?

acq4/devices/PatchPipette/states.py
```diff
@@ -1130,97 +1161,257 @@ class PatchPipetteResealState(PatchPipetteState):
...
+        if self._pressureFuture is not None:
+            self._pressureFuture.stop()
+        pos = np.array(self.dev.pipetteDevice.globalPosition())
+        # TODO move back a little?
+        pipette_direction = self.dev.pipetteDevice.globalDirection()
+
+        def random_wiggle_direction():
+            """pick a random point on a circle perpendicular to the pipette axis"""
+            while np.linalg.norm(vec := np.cross(pipette_direction, np.random.uniform(-1, 1, size=3))) == 0:
+                pass  # prevent division by zero
```
outofculture, line 1213: If you want to point out guards that are unlikely to happen, though, this is really very unlikely. The chance we could get two identical arrays of 3 floats is 1 in 55,340,232,221,128,654,848.

outofculture: oh, well, somewhat less for the constraint of `(-1, 1)`, but you get the idea.

acq4/devices/PatchPipette/states.py
campagnola: I'd assume the procedure looks like:
- nuzzle nucleus while slowly ramping up vacuum
- hold vacuum constant and slowly retract
- watch for changes in resistance

acq4/devices/PatchPipette/states.py
campagnola: I hope so :D

acq4/devices/PatchPipette/states.py
```diff
@@ -1130,97 +1163,219 @@ class PatchPipetteResealState(PatchPipetteState):
...
+            start = ptime.time()
+            while ptime.time() - start < self.config['nuzzleDuration']:
+                while np.dot(direction := random_wiggle_direction(), prev_dir) > 0:
+                    pass  # ensure different direction from previous
+                self.waitFor(self.dev.pipetteDevice._moveToGlobal(pos=pos + direction, speed=self.config['nuzzleSpeed']))
+                prev_dir = direction
+            self.waitFor(self.dev.pipetteDevice._moveToGlobal(pos=pos, speed=self.config['nuzzleSpeed']))
+            self.waitFor(self._pressureFuture)
+
+    def handleTear(self):
```
campagnola, line 1229: Maybe we don't need handleTear / handleStretch / tearRecoveryTime / stretchRecoveryTime ?

outofculture:
this is all old. we should just murge it.

PR #211, opened 2024-03-12
imports are now processed before devices

MoxenWolf:


acq4/Manager.py
```diff
@@ -288,15 +288,16 @@ def configure(self, cfg):
         self.sigConfigChanged.emit()
 
     def _loadConfig(self, cfg):
+        # Handle custom import prior to loading devices
+        if 'imports' in cfg:
+            if isinstance(cfg["imports"], str):
+                cfg["imports"] = [cfg["imports"]]
+            for mod in cfg["imports"]:
+                __import__(mod)
```
outofculture, line 296: This loses the `try` block wrapping. I'm personally okay with crashing the app on exceptions that happen during config-file processing (I always use `--exit-on-error`), but maybe double-check with Luke that he's okay never ignoring errors while custom importing.

campagnola: I think I'd prefer to follow the existing behavior of ignoring errors unless --exit-on-error is given.

PR #214, opened 2024-03-27
MultiPatch pipette control ui updates

campagnola:
Allow better control over clamp mode, holding, and auto bias.
- Change holding values regardless of current clamp mode
- Enable/disable bias regardless of mode
- Optionally lock bias to VC
- Easily disable bias by changing holding current
- Controls illuminate to show currently active control

Bonus: add a mock clamp simulator based on neuroanalysis, fix NEURON simulator


outofculture:
Superceded by #216

PR #216, opened 2024-04-06
MultiPatch pipette control ui updates

outofculture:
Allow better control over clamp mode, holding, and auto bias.

- Change holding values regardless of current clamp mode
- Enable/disable bias regardless of mode
- Optionally lock bias to VC
- Easily disable bias by changing holding current
- Controls illuminate to show currently active control

Bonus: add a mock clamp simulator based on neuroanalysis, fix NEURON simulato

(This takes over #214)

### DIFF
--- a/acq4/devices/PatchPipette/states.py
+++ b/acq4/devices/PatchPipette/states.py
@@ -158,7 +158,6 @@ class PatchPipetteState(Future):
                 cdev.setHolding(mode=mode, value=holding)
             if tpParams is None:
                 tpParams = {}
-            tpParams.setdefault('clampMode', mode)
 
         # enable test pulse if config requests it AND the device is "active"
         if tp is not None:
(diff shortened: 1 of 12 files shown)


acq4/devices/PatchPipette/states.py
```diff
@@ -158,7 +158,6 @@ def initializeClamp(self):
                 cdev.setHolding(mode=mode, value=holding)
             if tpParams is None:
                 tpParams = {}
-            tpParams.setdefault('clampMode', mode)
```
outofculture, line 161: This is the only bit of this PR I can't easily justify. Why don't we want to set the `clampMode` on the test pulse params, @campagnola ?

campagnola: Initial clamp mode should be set only once when the state begins; the test pulse then uses whatever mode is currently in use. (otherwise the test pulse is constantly enforcing a particular clamp mode, which we usually don't want)

PR #243, opened 2024-06-07
put position into range-of-motion error

### LINKED ISSUE #240: put saved position warning in terms of manipulator coordinates

outofculture:
closes #240

### DIFF
--- a/acq4/devices/Pipette/pipette.py
+++ b/acq4/devices/Pipette/pipette.py
@@ -198,10 +198,14 @@ class Pipette(Device, OptomechDevice):
         if bad_axes:
             axis_names = {0: 'x', 1: 'y', 2: 'z'}
             axes = ', '.join(axis_names[axis] for axis in bad_axes)
+            pos = self.mapGlobalToParent(pos)
             raise HelpfulException(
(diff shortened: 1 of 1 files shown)


acq4/devices/Pipette/pipette.py
```diff
@@ -198,10 +198,14 @@ def checkRangeOfMotion(self, pos, tolerance=500e-6):
         if bad_axes:
             axis_names = {0: 'x', 1: 'y', 2: 'z'}
             axes = ', '.join(axis_names[axis] for axis in bad_axes)
+            pos = self.mapGlobalToParent(pos)
```
campagnola, line 201: I don't think mapGlobalToParent gets you the coordinates you want here. Try following the position coordinate forward until it is actually sent to the stage device so you can see the transforms applied. I think it will involve Piptte._solveGlobalStagePosition and Stage.mapGlobalToDevicePosition.

outofculture: uh oh! that's what I'm using the _check_ the boundaries, too. okay, yeah, for a _manipulator_ to get to `pos` would be silly we need "when the pipette is at pos, what would its manipulator's position be?".

PR #244, opened 2024-06-11
Send the pipette home after cleaning

outofculture:


acq4/devices/Stage/Stage.py
```diff
@@ -695,11 +695,12 @@ def _movePath(self):
                     break
 
                 if fut.wasInterrupted():
```
campagnola, line 697: This makes me nervous -- are we just covering up some bad logic that occurred earlier?

outofculture: hmm, yeah, you're right to be nervous. I don't know what code path is getting us here! I was hoping this change would make the error more apparent, but it didn't. as it was, the error we were getting was "_taskDone called twice", which doesn't get us any details. with this change, though, we don't get any error, and then I forgot about it, which is definitely worse. REVERT!

Issue #246, opened 2024-06-11
Mosaic imaging fails with timeout


outofculture:
After collecting the first row of a mosaic:
```
  2024-06-11T13:54:02.536265-07:00 Unexpected error: 
  RuntimeError: Task <acq4.util.future.Future object at 0x000001E0D82B0820> did not complete: Timed out waiting for <acq4.util.future.MultiFuture object at 0x000001E0D8588550>
  RuntimeError: Task <acq4.util.future.Future object at 0x000001E0D82B0820> did not complete: Timed out waiting for <acq4.util.future.MultiFuture object at 0x000001E0D8588550>
    File C:\Users\svc_multipatch\acq4\acq4\util\future.py, line 225, in waitFor
      future.wait(0.1)
    File C:\Users\svc_multipatch\acq4\acq4\util\future.py, line 163, in wait
      raise self.Timeout("Timeout waiting for task to complete.")
  acq4.util.future.Future.Timeout: Timeout waiting for task to complete.
  
  The above exception was the direct cause of the following exception:
  
    File C:\Users\svc_multipatch\acq4\acq4\util\future.py, line 67, in executeAndSetReturn
      self._taskDone(returnValue=func(*args, **kwds))
                                 ^^^^^^^^^^^^^^^^^^^
    File C:\Users\svc_multipatch\acq4\acq4\util\imaging\sequencer.py, line 241, in run_image_sequence
      _future.waitFor(move)
    File C:\Users\svc_multipatch\acq4\acq4\util\future.py, line 229, in waitFor
      raise future.Timeout(f"Timed out waiting for {future!r}") from e
  acq4.util.future.Future.Timeout: Timed out waiting for <acq4.util.future.MultiFuture object at 0x000001E0D8588550>
  
  The above exception was the direct cause of the following exception:
  
    File C:\Users\svc_multipatch\acq4\acq4\util\imaging\sequencer.py, line 494, in threadStopped
      fut.wait(timeout=1) # to raise errors if any happened
      ^^^^^^^^^^^^^^^^^^^
    File C:\Users\svc_multipatch\acq4\acq4\util\future.py, line 180, in wait
      raise RuntimeError(msg) from self._excInfo[1]```

campagnola:
From my observations today, it looks like the sensapex hardware crashes (sometimes requiring multiple power cycles to get working again), which causes the timeout.

outofculture:
observation today:

- only happens on Y movement

outofculture:
[...]

PR #248, opened 2024-06-12
Put a thread dump into the log window for exceptions

outofculture:
helps us track down #246

### DIFF
--- a/acq4/util/LogWindow.py
+++ b/acq4/util/LogWindow.py
@@ -717,6 +717,12 @@ class LogWidget(Qt.QWidget):
                     e["tracebackHtml"],
                     doc,
                 )
+            if "threadsHtml" in e:
+                doc = re.sub(
+                    f'<a href="threads:{e["id"]}">(<[^>]+>)*Show thread states {e["id"]}(<[^>]+>)*</a>',
+                    e["threadsHtml"],
+                    doc,
+                )
 
         with open(fileName, "wb") as f:
             f.write(doc.encode("utf-8"))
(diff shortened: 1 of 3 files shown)


acq4/util/LogWindow.py
```diff
@@ -677,6 +670,18 @@ def formatTracebackForHTML(self, tb):
             cleanLines.append(line)
         return ''.join(cleanLines)
 
+    def formatThreadsForHTML(self, entry):
+        threads = entry["threads"]
+        traceback.format_stack
```
campagnola, line 675: ?

acq4/util/debug.py
```diff
@@ -4,10 +4,11 @@
 Copyright 2010  Luke Campagnola
 Distributed under MIT/X11 license. See license.txt for more information.
 """
-from __future__ import print_function
 
-from pyqtgraph.debug import *
+import sys
+
 import pyqtgraph.debug as pgdebug
+from pyqtgraph.debug import *
```
campagnola, line 11: Better to pick a specific list of names to import?

outofculture: I agree! The actual use this is getting though is that a bunch of other places are `from acq4.debug import THING_IMPORTED_FROM_PG`, and I didn't want to track down all of those.

outofculture: (I was just moving imports around in this file)

campagnola: Ahaaaa, yeah it does look more like something I would do.

PR #252, opened 2024-06-20
Stage transforms should only translate

outofculture:


acq4/devices/Stage/Stage.py
```diff
@@ -316,6 +316,16 @@ def _updateTransform(self):
         ## this informs rigidly-connected devices that they have moved
         self.setDeviceTransform(self._baseTransform * self._stageTransform)
 
+    def setDeviceTransform(self, tr):
+        is_dict = isinstance(tr, dict)
+        super().setDeviceTransform(tr)
+        if is_dict:  # only do this during init, basically
```
campagnola, line 322: If this is only during init, then why not check this inside `__init__` where we know the transform is defined by a dict with pos/scale/angle keys?

PR #255, opened 2024-06-25
presets for microscopes

### LINKED ISSUE #253: teach Microscope to have presets
put it in the config.

* teach microscope to discover all its light sources
* give all devices a chance to do "post-device setup" setup ("device-configured" signal? only children?)
* teach it to know its filters
* give optomech "optics" (e.g. transmitted, epifluorescent) labels on parent-child relationships (similar to filterset)
* maybe make a context manager that preserves previous states once the context is done

outofculture:
closes #253

### DIFF
--- a/acq4/devices/VimbaXCamera.py
+++ b/acq4/devices/VimbaXCamera.py
@@ -285,10 +285,11 @@ def _featureNameToParamName(name):
 
 
 def main():
-    class MockManager:
-        def declareInterface(self, *args, **kwargs):
-            pass
-    cam = VimbaXCamera(MockManager(), {'id': 'DEV_000F315B9827'}, 'test')
+    import pyqtgraph as pg
+    from acq4.Manager import Manager
+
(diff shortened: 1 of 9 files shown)


acq4/devices/VimbaXCamera.py
```diff
@@ -285,10 +291,12 @@ def _featureNameToParamName(name):
...
-    class MockManager:
-        def declareInterface(self, *args, **kwargs):
-            pass
-    cam = VimbaXCamera(MockManager(), {'id': 'DEV_000F315B9827'}, 'test')
+    import pyqtgraph as pg
+    from acq4.Manager import Manager
+
+    pg.mkQApp()  # for event loop
+    id_ = VimbaXCamera.listCameras()[0]
+    cam = VimbaXCamera(Manager(), {'id': id_}, 'test')
```
outofculture, line 299: okay, so the refactors of Manager above are to make it so I can instantiate one here in a test.

PR #260, opened 2024-07-01
Sequencer stop button enhancement

outofculture:


acq4/util/future.py
```diff
@@ -220,7 +228,11 @@ def waitFor(self, future: Future[WAITING_RETVAL_TYPE], timeout=20.0) -> Future[W
         """
         start = time.time()
         while True:
-            self.checkStop()
+            try:
+                self.checkStop()
+            except self.StopRequested:
+                future.stop(reason="parent task stop requested")
```
outofculture, line 234: this try/except wasn't actually necessary to get the correct behavior, but it seems like the right thing to do.

acq4/util/imaging/sequencer.py
```diff
@@ -240,7 +240,7 @@ def handle_new_frames(f: "Frame | list[Frame]", idx: int):
                 for move in movements_to_cover_region(imager, mosaic):
                     _future.waitFor(move)
                     if z_stack:
-                        stack = acquire_z_stack(imager, *z_stack, block=True).getResult()
```
outofculture, line 243: @campagnola here's the problematic call. we can't make it a separate thread, or the device locking won't work. we can't rely on `stop()` on the parent, since it's not in a `checkStop` loop at this point.

PR #261, opened 2024-07-02
Add new vid:pid for scientifica controllers

campagnola:
WIP; this code needs to be tested.

### DIFF
--- a/acq4/drivers/Scientifica/scientifica.py
+++ b/acq4/drivers/Scientifica/scientifica.py
@@ -69,17 +69,22 @@ class Scientifica(SerialDevice):
         Sets Scientifica.availableDevices to a dict of {name: port} pairs.
 
         This works by searching for USB serial devices with device IDs used by Scientifica
-        (vid=0403, pid=6010) and sending a single serial request.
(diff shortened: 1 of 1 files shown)


acq4/drivers/Scientifica/scientifica.py
```diff
@@ -69,17 +69,22 @@ def enumerateDevices(cls) -> dict[str, str]:
...
-            if ('VID_0403' not in ident or 'PID_6010' not in ident) and '0403:6010' not in ident:
+            isScientifica = False
+            for vid,pid in [('0403', '6010'), ('10C4', 'EA70')]:
+                # several different ways this can appear:
+                #  VID_0403+PID_6010
+                #  VID_0403&PID_6010
+                #  VID:PID=0403:6010
+                if (f'VID_{vid}' in ident and f'PID_{pid}' in ident) or f'{vid}:{pid}' in ident:
+                    isScientifica = True
+                    break
```
outofculture, line 86: ~This `break` is going to skip all the rest of the `for` loop, which includes actually instantiating the object and adding it to `devs`.~ Nvm; you added an inner loop.

outofculture (approved):

PR #264, opened 2024-07-12
Avoid obstacles during 'cell detect'

### LINKED ISSUE #263: Obstacle avoidance during cell detection
When approaching a cell through tissue, the operator usually watches the pipette tip for contact with any obstacles hit before reaching the cell. This is most apparent when the tip is blocked, causing the pressure wave to dissipate from the surrounding tissue. For automation, though, it is easier to just watch for changes in resistance. So, in cell-detect mode:
 - Advance toward the cell while monitoring resistance
 - If resistance increases beyond a threshold before reaching the cell, move back about 10 μm and check that the resistance recovered (if it doesn't, then the tip is fouled)
 - Move the pipette 5-10 μm laterally to avoid the obstacle, then advance 20 μm and return to the original axial path to the target

outofculture:
closes #263

### DIFF
--- a/acq4/devices/PatchPipette/states.py
+++ b/acq4/devices/PatchPipette/states.py
@@ -239,7 +239,10 @@ class PatchPipetteState(Future):
     def __repr__(self):
         return f'<{type(self).__name__} "{self.stateName}">'
 
-    def wiggle(self, speed, radius, repetitions, duration, extra=None):
+    def wiggle(self, speed, radius, repetitions, duration, pipette_direction=None, extra=None):
+        if pipette_direction is None:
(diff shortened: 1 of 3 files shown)


acq4/devices/PatchPipette/states.py
```diff
@@ -236,6 +239,51 @@ def checkStop(self, delay=0):
     def __repr__(self):
         return f'<{type(self).__name__} "{self.stateName}">'
 
+    def wiggle(self, speed, radius, repetitions, duration, pipette_direction=None, extra=None):
```
outofculture, line 242: move this onto Pipette

PR #282, opened 2024-09-12
Changes to cell detect, seal, reseal

outofculture:


acq4/util/LogWindow.py
```diff
@@ -312,7 +312,7 @@ def setLogDir(self, dh: DirHandle):
             if dh.exists("log.txt"):
                 self.logFile = dh["log.txt"]
                 try:
-                    tempCount = max(e["id"] for e in configfile.readConfigFile(self.logFile.name()).values())
+                    tempCount = max(e["id"] for e in configfile.readConfigFile(self.logFile.name()).values()) + 1
```
outofculture, line 315: I couldn't figure out why this was off by one, but the new pyqtgraph conffile error-on-duplicate-keys behavior was getting triggered without this.

PR #284, opened 2024-09-17
implement row-major as the image default

outofculture:
depends on #285


outofculture:
need to make sure coordinate access of images is correct:
 - scanner
 - object detection
 - mosaic
 - make sure frame transforms are correct
 - saved images load their transform properly (saved pre- and post-this-change)

PR #285, opened 2024-09-19
Start toward deprecating analysis

outofculture:


outofculture:
it's impossible to know if anyone else is using any of these files/classes/functions, but the ACQ4 usages I could find work after these moves.

PR #296, opened 2024-10-10
IN PROGRESS: 3D visualization

outofculture:


acq4/devices/Device.py
```diff
@@ -56,7 +59,23 @@ def deviceInterface(self, win):
...
+
+    def getGeometries(self) -> list:
+        """Return a 3D model to be displayed in the 3D visualization window.
+        """
+        from acq4.modules.Visualize3D import create_geometry
+
+        defaults = self.defaultGeometryArgs()
+        if self._geometry:
+            args = {**self._geometry}
+            return create_geometry(defaults=defaults, **args)
```
outofculture, line 76: It would be nice if we could apply these defaults deeply. E.g. setting the rotation on the transform without overriding the entire transform.

PR #312, opened 2025-01-24
Clean up locking in Manager

campagnola:


outofculture (approved):
that should clear up my problems, yeah.

PR #313, opened 2025-01-24
log whenever the patching config gets changed

outofculture:


acq4/modules/MultiPatch/patchProfileEditor.py
```diff
@@ -48,6 +51,7 @@ def paramTreeChanged(self, root_param, changes):
                         continue
                     if PatchPipetteStateManager.getProfileConfig(profile_item.name()).get("copyFrom", None) == profile_name:
                         profile_item.applyDefaults({state_name: {param_name: data}})
+        logMsg(f"Pipette profile {profile_name} updated: {json.dumps(changes, cls=ACQ4JSONEncoder)}")
```
campagnola, line 54: I imagined this being recorded in the multipatch log so that we can ask it what parameters were in use at any given time point.

outofculture: the two logs both have timestamps, so it wouldn't be too hard to correlate them, but I can see that it would be convenient if they were all in one place. I'll add it there, too, as well as at the beginning of any multipatch log.

PR #314, opened 2025-01-31
(disabled by default) new pipette motion planning

### LINKED ISSUE #187: Use actual 3D motion planning / collision detection
Recommend Klamp't: robotics / IK / motion planning / control library. Looks like it has everything we would need, including PyQt visualizations in opengl. Another option is pymp (https://github.com/Jiayuan-Gu/motion-planning/tree/main)

Would be an ideal approach to solving #161, but will take more time

* [x] Declare geometry of optomech devices in config (mostly boolean combinations of truncated cones)
* [x] Module for visualizing geometry in 3D (maybe vispy, maybe other..)
* [x] Visualization should handle changes in geometry
* [x] Get all the transforms right (pipette!)
* [x] Motion planner that automatically avoids collisions
* [x] Sample surface handling
* [x] Optimizations
* [ ] Bonus: Motion planning system that can look into the future and avoid collisions even when the scene is changing over time

outofculture:
Instead of simple heuristics, let's use a fully equipped motion planner! Mostly closes #187


acq4/devices/geometries/sensapex_lefthanded.srdf
```diff
@@ -0,0 +1,34 @@
+<robot name="linear_actuator_robot">
```
campagnola, line 1: Maybe remove these? Or collect all of the almost-pinocchio stuff into one place?

config/example/devices.cfg
```diff
@@ -60,28 +60,93 @@ Clamp1:
...
+    vcHolding: -65e-3
+
 
 # RecordingChamber is used to mark the position and boundaries of a recording chamber under the microscope.
 # In some setups, these are used to inform automated patch planning.
+
 RecordingChamber:
     driver: 'RecordingChamber'
-    radius: 5*mm
+    radius: 11*mm
```
campagnola, line 90: probably don't need this anymore?

config/example/devices.cfg
```diff
@@ -60,28 +60,93 @@ Clamp1:
...
+    # Default holding values.
+    icHolding: 0.0
+    vcHolding: -65e-3
+
 
 # RecordingChamber is used to mark the position and boundaries of a recording chamber under the microscope.
 # In some setups, these are used to inform automated patch planning.
+
 RecordingChamber:
     driver: 'RecordingChamber'
```
campagnola, line 89: Wondering if we need a bare Geometry device class

acq4/devices/PatchPipette/states/clean.py
```diff
@@ -19,9 +19,6 @@ class CleanState(PatchPipetteState):
     rinseSequence : list
         List of (pressure (Pa), duration (s)) pairs specifying how to pulse pressure while the pipette tip is in the
         rinse well.
-    approachHeight : float
```
campagnola, line 22: We still need approachHeight.. move it to Pipette's config if it doesn't work here.

acq4/__main__.py
```diff
@@ -174,5 +174,5 @@ def save_history(historyPath=historyPath):
             app.exec_()
     else:
         app.exec_()
-        # pg.exit()  # pg.exit() causes python to exit before Qt has a chance to clean up. 
+        pg.exit()  # pg.exit() causes python to exit before Qt has a chance to clean up.
```
outofculture, line 177: I uncommented this line because without it, every time I exit the app I get an ugly error:
```
Qt Error: (traceback follows)
QThread: Destroyed while thread is still running
  File "/home/martin/src/acq4/acq4/acq4/__main__.py", line 52, in messageHandler
    traceback.print_stack()
Fatal error occurred; asking manager to quit.
Fatal Python error: Aborted

Thread 0x000072cdb492b640 (most recent call first):
  File "/home/martin/.miniforge3/envs/acq4-torch/lib/python3.11/threading.py", line 327 in wait
  File "/home/martin/.miniforge3/envs/acq4-torch/lib/python3.11/queue.py", line 171 in get
  File "/home/martin/src/acq4/pyqtgraph/pyqtgraph/console/repl_widget.py", line 182 in run

Thread 0x000072cdb67fc640 (most recent call first):
  File "/home/martin/.miniforge3/envs/acq4-torch/lib/python3.11/threading.py", line 327 in wait
  File "/home/martin/.miniforge3/envs/acq4-torch/lib/python3.11/queue.py", line 171 in get
  File "/home/martin/src/acq4/acq4/acq4/modules/Visualize3D/travelers_proxy.py", line 101 in _watchForPathUpdates
  File "/home/martin/.miniforge3/envs/acq4-torch/lib/python3.11/threading.py", line 982 in run
  File "/home/martin/.miniforge3/envs/acq4-torch/lib/python3.11/threading.py", line 1045 in _bootstrap_inner
  File "/home/martin/.miniforge3/envs/acq4-torch/lib/python3.11/threading.py", line 1002 in _bootstrap

Thread 0x000072cdec9fd640 (most recent call first):
  File "/home/martin/.miniforge3/envs/acq4-torch/lib/python3.11/threading.py", line 331 in wait
  File "/home/martin/.miniforge3/envs/acq4-torch/lib/python3.11/queue.py", line 180 in get
  File "/home/martin/src/acq4/acq4/acq4/devices/Camera/Camera.py", line 867 in run
  File "/home/martin/src/acq4/acq4/acq4/util/Thread.py", line 34 in __run_wrapper

Thread 0x000072cded1fe640 (most recent call first):
  File "/home/martin/src/acq4/acq4/acq4/devices/FilterWheel/filterwheel.py", line 461 in run
  File "/home/martin/src/acq4/acq4/acq4/util/Thread.py", line 34 in __run_wrapper

Current thread 0x000072cf24a37440 (most recent call first):
  <no Python frame>
```

acq4/devices/Device.py
```diff
@@ -32,6 +32,7 @@ def __init__(self, deviceManager: acq4.Manager.Manager, config: dict, name: str)
         self._lock_tb_ = None
         self.dm = deviceManager
         self.dm.declareInterface(name, ['device'], self)
+        self._config = config
```
outofculture, line 35: I feel like I've been told not to do this, that it's "here be dragons" territory, but it behaves well for all the hardware we've tested it on, so... maybe we call it `_untrrustworthyConfig` or some such, so we won't feel as bad later when it somehow fails?

acq4/devices/Pipette/planners.py
```diff
@@ -183,8 +335,9 @@ def stop(self):
         if self.future is not None:
             self.future.stop()
 
-    def _move(self):
-        return self.pip._movePath(self.path())
+    @future_wrap
+    def _move(self, _future):
+        _future.waitFor(self.pip._movePath(self.path()))
```
outofculture, line 340: this seems wrong

acq4/util/Qt.py
```diff
@@ -182,3 +182,46 @@ def _do_layout(self, rect, test_only):
             line_height = max(line_height, item.sizeHint().height())
 
         return y + line_height - rect.y()
+
+
+orig_QObject = QObject
+
+
+class QObject(orig_QObject):
```
outofculture, line 190: this all goes away in patch-me-if-you-can, so we can ignore it

acq4/devices/Device.py
outofculture: searching around, this line is also present in Microscope, Filter, ThorLab's Filter, and VimbaXCamera (I added that). there are lots of weird `self.config =` choices to be wary of, though. maybe that's all I was warned off of? okay, but then that makes this a shitty proliferation of duplicate ways to access the config.

outofculture: oh, we also have `self.__config =` in Optomech.

PR #316, opened 2025-02-07
dev from rig 5

ben-at-allen:
Work from Luke and I on rig 5.

### DIFF
--- a/acq4/__main__.py
+++ b/acq4/__main__.py
@@ -39,6 +39,9 @@ from .util import Qt
 
 # Import pyqtgraph, get QApplication instance
 import pyqtgraph as pg
+if "--threadtrace" in sys.argv:
+    sys.argv.pop(sys.argv.index('--threadtrace'))
+    tt = pg.debug.ThreadTrace()
 
 app = pg.mkQApp()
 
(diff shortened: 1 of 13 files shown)


acq4/__main__.py
```diff
@@ -39,6 +39,9 @@
 
 # Import pyqtgraph, get QApplication instance
 import pyqtgraph as pg
+if "--threadtrace" in sys.argv:
+    sys.argv.pop(sys.argv.index('--threadtrace'))
+    tt = pg.debug.ThreadTrace()
```
outofculture, line 44: putting a ticket in to use argparse so we can get these args documented

acq4/devices/PatchPipette/statemanager.py
```diff
@@ -88,15 +88,15 @@ def getProfileConfig(cls, name):
             raise KeyError(f"Unknown patch profile {name}")
         return cls.profiles[name]
 
-    @classmethod
-    def _loadGlobalProfilesOnce(cls):
-        if cls._profilesLoadedFromConfig:
+    @staticmethod
```
outofculture, line 91: this sure was a fun problem to find :D

outofculture (approved):
I can't say I know anything about igor or mies, but nothing in this PR raises any red flags for me.

acq4/devices/PatchPipette/statemanager.py
campagnola: Thanks for solving the mystery :)

PR #322, opened 2025-03-04
Pipette detection model 8

campagnola:
Pipette detection updates; working nicely with model 8


acq4/util/imaging/object_detection.py
```diff
@@ -16,6 +16,8 @@
 from pyqtgraph import SRTTransform3D
 from teleprox import ProcessSpawner
 from teleprox.shmem import SharedNDArray
+import coorx.image
```
outofculture, line 19: we'll need to release a new version of coorx

outofculture (approved):

PR #332, opened 2025-03-19
Sonicator support

### LINKED ISSUE #328: Add a sonicator device base class + DAQ-driven subclass
The base really only needs one public method: `sonicate(mode)` that should return a future, where `mode` could be `clean`, `expel`, or others. The primary reason to keep this API very simple is that we might want to support other sonication hardware in the future and we don't know exactly what features they will support.

The DAQ subclass should have config options that define what happens during each sonication mode -- a sequence of (frequency, amplitude, duration) triples, similar to the pipette cleaning config.

PatchPipette now gets an optional 4th subdevice and the `clean` and `expel` states may activate the sonicator.

outofculture:
closes #328


acq4/devices/DAQSonicator.py
```diff
@@ -0,0 +1,80 @@
...
+            voltage = self.calcVoltage(frequency)
+            wave = Sine(0, duration, frequency, voltage).eval(sample_rate=self._daq.sampleRate).data
+            numPts = len(wave)
+            cmd = {
+                "protocol": {"duration": duration},
+                self._daq.getDAQName("analog"): {
+                    "rate": 500_000,
+                    "numPts": numPts,
+                },
+                self._daq.name(): {
```
campagnola, line 55: No waveform needed; just set the holding to 1 during and 0 after
This will especially help with devices like yours that can't to time-sampled DIO

acq4/devices/DAQSonicator.py
```diff
@@ -0,0 +1,80 @@
...
+                    "analog": {"command": wave},
+                    "digital": {"command": np.ones(numPts)},
+                },
+            }
+
+            task = self.dm.createTask(cmd)
+            _future.checkStop()
+            task.execute()
```
campagnola, line 63: We need to wait for the result while running checkStop in a loop?

acq4/devices/PatchPipette/states/nucleus_collect.py
```diff
@@ -49,6 +55,11 @@ def run(self):
         # self.waitFor([pip._moveToGlobal(self.approachPos, speed='fast')])
         self.waitFor(pip._moveToGlobal(self.collectionPos, speed='fast'), timeout=None)
 
+        if dev.sonicatorDevice is not None:
+            dev.sonicatorDevice.doProtocol(
+                'expel', duration=config['sonicationDuration'], frequency=config['sonicationFrequency']
+            ).raiseErrors("Error sonicating pipette")
```
campagnola, line 61: Keep the future and wait on it later

acq4/devices/Sonicator.py
```diff
@@ -0,0 +1,210 @@
...
+
+    sigSonicationChanged = Qt.pyqtSignal(float)
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.actionLock = RLock()
+
+    @future_wrap
+    def doProtocol(self, mode: str, _future: Future = None, **kwargs) -> None:
+        with self.actionLock:
```
campagnola, line 22: Modes should be defined in config:
```
duration = kwds.get('duration', self.configuredModes[mode]['duration'])
```

acq4/devices/Sonicator.py
```diff
@@ -0,0 +1,210 @@
...
+                    _future.waitFor(self.sonicate(frequency, step_duration, lock=False))
+                    frequency += step
+                frequency -= step
+                while frequency > start:
+                    _future.waitFor(self.sonicate(frequency, step_duration, lock=False))
+                    frequency -= step
+            else:
+                raise ValueError(f"Unrecognized sonication protocol '{mode}'")
+
+    def sonicate(self, frequency: float, duration: float, lock: bool = True) -> Future:
```
campagnola, line 43: Use device reservation instead of passing a lock around? And in the subclass, make sure to reserve both devices together.

outofculture: I think it is enough to just reserve the daq generic device (and the daq). I've removed other locks.

PR #338, opened 2025-04-01
use previous tip positions to set new pipette position

### LINKED ISSUE #327: Keep track of average pipette tip position
- For each pipette device, store a history of the last ~20 calibrated pipette positions (and the date of the calibration)
- If the pipette is calibrated to an unexpected location (outside 3 σ or 1 mm on any axis?) then pop up a message asking whether there is a mistake (cancel the calibration; operator needs to fix something and calibrate again), or this is a temporary fluke to accept (keep the calibration but do not add it to the history), or this is the new norm (clear the history before this point).
- when auto calibrating a new pipette, reset the calibration to the average before proceeding

outofculture:
Closes #327


acq4/devices/Pipette/pipette.py
```diff
@@ -230,6 +233,97 @@ def cameraModuleInterface(self, mod):
         self._camInterfaces[iface] = None
         return iface
 
+    def tipOffsetIsReasonable(self, pos) -> bool:
+        dist = np.linalg.norm(np.array(self.mapToGlobal((0, 0, 0))) - pos)
+        return dist < 30e-6
```
campagnola, line 238: this threshold should be configurable

PR #350, opened 2025-04-18
offer config option to change the definition of "reasonable"

outofculture:


acq4/devices/Pipette/pipette.py
```diff
@@ -69,6 +69,9 @@ class Pipette(Device, OptomechDevice):
     * idleDistance: the x/y distance from the global origin from which the pipette top should be placed
       in idle mode. Default is 7 mm.
     * recordingChambers: list of names of RecordingChamber devices that this Pipette is meant to work with.
+    * reasonable tip offset distance: the maximum distance (in meters) from the original tip offset that is
```
outofculture, line 72: I am frustrated that we have this tension of "users expect camel case" throughout our app. It's a worse user interface, to write words intended for humans to read without spaces. Fixing it everywhere all at once is too expensive (maybe?). Fixing it in a few places will lead to confusion, and we should only do so if our eventual goal is to change it everywhere. That would be my preference.

campagnola: I am ok with fixing all at once, but maybe there's  conversation to be had about underscores vs spaces?

campagnola: For now, though, let's go for consistency.

PR #360, opened 2025-05-07
add tool for using claude to search through data directories

outofculture:
depends on #314


campagnola:
Too dangerous to use, too weird to die.

PR #362, opened 2025-05-09
for moving very slow, we also need to set the minSpeed

outofculture:


outofculture:
this doesn't work in practice

PR #363, opened 2025-05-13
changes from rig 5

ben-at-allen:
These are the changes from rig 5 that I had a part in.  I remove some of the superfluous comments and added in a raised exception.  I will have some time available tomorrow again to address any needed changes.

### DIFF
--- a/acq4/modules/MultiPatch/pipetteControl.py
+++ b/acq4/modules/MultiPatch/pipetteControl.py
@@ -43,15 +43,8 @@ class PipetteControl(Qt.QWidget):
             if self.pip.clampDevice is not None:
                 self.pip.clampDevice.sigTestPulseFinished.connect(self.updatePlots)
                 self.pip.clampDevice.sigAutoBiasChanged.connect(self._updateAutoBiasUi)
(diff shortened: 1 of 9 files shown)


campagnola:
Other than a few more debugging prints that should be removed, it looks great!

acq4/modules/MultiPatch/pipetteControl.py
```diff
@@ -43,15 +43,8 @@ def __init__(self, pipette: PatchPipette, mainWin, parent=None):
             if self.pip.clampDevice is not None:
                 self.pip.clampDevice.sigTestPulseFinished.connect(self.updatePlots)
                 self.pip.clampDevice.sigAutoBiasChanged.connect(self._updateAutoBiasUi)
-            if self.pip.pressureDevice is None:
```
campagnola, line 46: It looks like some code got incorrectly overwritten in this file?

ben-at-allen: @campagnola ok this should be fixed now.  I have no idea how that happened :o

PR #365, opened 2025-05-21
Monster change storm toward full automation

outofculture:


acq4/devices/PatchPipette/states/reseal.py
```diff
@@ -364,7 +364,7 @@ def run(self):
             self.sleep(0.2)
 
         self.setState("reseal deemed successful")
-        self.cleanup()
+        self._cleanup()
```
outofculture, line 367: this maybe shouldn't change.

PR #382, opened 2025-07-21
Support for single-frame feature tracking

outofculture:


acq4/devices/Camera/Camera.py
```diff
@@ -444,6 +444,21 @@ def getBoundary(self, globalCoords: bool = True, mode="sensor") -> tuple:
         else:
             return bounds
 
+    def pointIsVisible(self, point, mode="sensor") -> bool:
```
campagnola, line 447: Should check that point is only 2D

acq4/devices/PatchPipette/states/_base.py
```diff
@@ -306,25 +324,67 @@ def depthBelowSurface(self, pos=None):
     def aboveSurface(self, pos=None):
         return self.depthBelowSurface(pos) < 0
 
+    def maybeVisuallyTrackTarget(self):
```
campagnola, line 327: Idly wondering if there are ways to prevent the base class becoming more complex in order to support features only used by a few states.

acq4/modules/AutomationDebug/cell.py
```diff
@@ -26,27 +27,27 @@ def __init__(self, position):
...
         self._trackingFuture = None
-        self.isTracking = False
         self._tracker = None
-        self._roiSize = None
 
     @property
-    def position(self):
-        """Get the current position of the cell."""
-        return np.array(self._positions[max(self._positions)])
+    def isTracking(self):
```
campagnola, line 33: Can all visual tracking API like isTracking and the signals above just be reached from `cell.tracker` ?

PR #383, opened 2025-07-25
Make device-device signal connections direct.

campagnola:
This moves stage update processing out of the GUI thread.

Also snuck in some minor mosaiceditor fixes


outofculture:
this has so far behaved well on a rig. tested with simultaneous device movements, camera operations, UI interactions.

PR #392, opened 2025-09-03
use logging and LogViewer

outofculture:


acq4/Manager.py
```diff
@@ -659,41 +664,60 @@ def getCurrentDir(self):
...
+        self._logFile = d["log.json"]
+        file_handler = setup_logging(
+            self._logFile.name(),
+            root_level=self._rootLogLevel,
+            console_level=self._consoleLogLevel,
+        )
+        self.sigLogDirChanged.emit(d)
+        if was_temp:
+            try:
+                with open(TEMP_LOG, 'r') as f:
```
outofculture, line 681: just append the files (using DirHandle)

PR #413, opened 2025-09-29
optimization: rate limit Sensapex position updates inline

outofculture:
the update timer was seemingly out of control. this change assumes incidental UMP position updates are happening constantly at the <100nm level, and so no explicit timer is needed to ensure the position update will be revisited in a timely fashion.

### DIFF
--- a/acq4/devices/Sensapex.py
+++ b/acq4/devices/Sensapex.py
@@ -216,7 +216,6 @@ class Sensapex(Stage):
         Sensapex.devices.pop(self.devid, None)
         if len(Sensapex.devices) == 0:
             UMP.get_ump().close()
-        self._positionWatcher.join()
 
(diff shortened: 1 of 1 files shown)


acq4/devices/Sensapex.py
```diff
@@ -205,14 +198,10 @@ def _positionChanged(self, dev, newPos, oldPos):
...
         # rate limit updates to 10 Hz
-        wait = 100e-3 - (now - self._lastUpdate)
-        if wait > 0:
-            self._sigRestartUpdateTimer.emit(wait)
-        else:
-            self._getPosition()
-
-    def _restartUpdateTimer(self, wait):
-        self._updateTimer.start(int(wait * 1000))
+        if now - self._lastUpdate < 100e-3:
```
campagnola: This discards up to 100 ms of updates from the end of every move?

PR #446, opened 2025-11-13
feat: add automatic version incrementing on main branch commits

outofculture:
Adds a GitHub Actions workflow that automatically increments the patch version number whenever code is pushed to main (via PR merge or direct commit). The workflow also creates corresponding git tags.


outofculture:
github doesn't offer hooking into right before a PR merge commit is created, so the only way to get the version into that commit is to allow force-pushing on main, which is not a permission I want to grant to a bot just yet.

PR #504, opened 2026-04-01
Logging: bugfixes and improvements

outofculture:
## Summary

Refactors log initialization so that logging starts immediately when acq4 launches, before the Manager or config are set up, using a temporary log file that gets migrated once the real log directory is known.

Key changes:

- **`__main__.py`**: Calls `setup_logging()` early (with `is_temp_file=True`) right after the Qt app is created. The \"Loading ACQ4...\" message now goes through the logger instead of `print()`. Added `--console-log-level` and `--root-log-level` CLI args (moved from Manager). Adds stdout/stderr UTF-8 encoding reconfiguration for Windows.
- **`logging_config.py`**: Extracts file handler management into a new `set_log_file()` function and a `rewrite_log_from_temp_file()` helper. The file handler is now tracked as a module-level global (`log_file_handler`) with an `is_temp_file` flag, so migration from temp → real log file happens automatically when `set_log_file()` is called. `setup_logging()` no longer returns a file handler.
- **`Manager.py`**: Removes the early `setup_logging()` call and `TEMP_LOG` constant (moved to `__main__`). Removes `--log-level`/`--root-log-level` args and the `_consoleLogLevel`/`_rootLogLevel` instance vars. `setLogDir` is simplified to a single `set_log_file()` call. Shutdown log messages promoted from DEBUG to INFO.
- **`Interfaces.py`**: Minor — adds a clarifying comment to `sigInterfaceListChanged`.

## Test plan
- [ ] Start acq4 and confirm log output appears immediately (before config loads)
- [ ] Confirm early log messages (e.g. "Loading ACQ4...") appear in the final log file after startup
- [ ] Verify `--console-log-level` and `--root-log-level` args work from the command line
- [ ] Confirm no locked-file errors on Windows during startup or shutdown
- [ ] Confirm shutdown sequence logs module/device names at INFO level


outofculture (commented):
Thanks for the logging improvements. A few issues worth addressing before merge:

1. **`logging_config.py` ~line 19 — `__reload__` reads a stale key**

   `__reload__` calls `old.get('_handlers', [])`, but the variable was renamed to `log_handlers`. On hot-reload this means all active handlers are silently dropped rather than cleaned up. Change the key to `'log_handlers'`. Additionally, `log_file_handler` is never preserved across reloads, so it will also be lost on any config reload.

2. **`Manager.py` ~lines 103, 662, 680 — `self._logFile` never set; log-dir dedup broken**

   `setLogDir` previously set `self._logFile`, but this PR removed that assignment while leaving `self._logFile = None` in `__init__` and the read at ~line 680. As a result, `logDir` is always `None`, and `setCurrentDir` will call `setLogDir` on every invocation even when the directory hasn't changed — defeating the dedup logic entirely.

3. **`__main__.py` ~lines 33–35 — `reconfigure()` crashes when `stdin` is `None`**

   `sys.stdin.reconfigure(encoding='utf-8')` raises `AttributeError` (or `io.UnsupportedOperation`) when running under `pythonw.exe` on Windows (where `sys.stdin` is `None`) or when streams are replaced by test frameworks. This needs a `hasattr` / `is not None` guard before calling `reconfigure`.

4. **`logging_config.py` ~lines 199–205 — Commented-out GUI log backfill block**

   There is a block commented out that previously populated the GUI log window with migrated records. If this is intentional (because the GUI handler receives records live and backfill is no longer needed), please delete the dead code entirely. If it is a bug (early records won't appear in the GUI), restore it. Either way, commented-out code should not be merged.

5. **`Manager.py` ~line 28 — Two dead imports**

   `HistoricLogRecord` and `setup_logging` are imported but are no longer used anywhere in `Manager.py` after this refactor. These should be removed.

6. **`__main__.py` ~line 61 — Temp log file written to bare relative path**

   `"temp_log.json"` is created in the current working directory. Two concurrent acq4 processes launched from the same directory will race to write and delete the same file. (The presence of `temp_log.json` already checked into the repo root suggests this has already happened in practice.) Please use `tempfile.gettempdir()` or a `tempfile.NamedTemporaryFile` to give each process its own file.

PR #507, opened 2026-04-01
feat: add pipette event log widget with live and replay support

outofculture:
## Summary

- **New `PipetteEventLog` widget** (`acq4/util/ui/pipetteEventLog.py`): a scrollable `QListWidget` that displays a human-readable timeline of meaningful pipette state changes (state transitions, moves, calibration, tip cleanliness, etc.). In live mode it auto-scrolls to the latest event; scrolling up enters history mode. Clicking an entry emits `sigTimeSelected` so other widgets can seek to that time.
- **In-memory event log on `PatchPipette`**: introduces `LOG_EVENT_TYPES` (a frozenset of semantically meaningful event types) and a `_logEvents` list that accumulates matching events since the last `newPatchAttempt()`. A new `eventLog()` accessor exposes the list for UI initialization. The event log is cleared when `newPatchAttempt()` is called or when "Reset History" is clicked.
- **Live MultiPatch UI integration** (`pipetteControl.py`): each `PipetteControl` now contains a `PipetteEventLog` embedded in the clamp widget area. It subscribes to `sigNewEvent` and populates itself with any already-logged events when constructed.
- **MultiPatch UI layout refactor** (`multipatchTemplate.ui`, `multipatchTemplate.py`): the top-level layout is replaced with a `QSplitter` containing three panels — a left controls panel, the existing pipette matrix, and a new right plots panel. Per-pipette plots (formerly embedded per-row in the pipette template) are moved to the shared right panel. Splitter sizes are saved and restored with the window config. A bug in the `_shouldSaveTipImages` setter (always passing `True` instead of `value`) is fixed. A bug in `cameraModuleClicked_setTip` / `cameraModuleClicked_setTarget` (using `.z` attribute instead of `[2]` index on a position array) is fixed.
- **`MultiPatchLog` replay support**: `process()` now classifies log-event-type records under a `log_events` use, and `MultiPatchLogWidget` gains a `PipetteEventLog` panel that is populated from all devices' `log_events` on load. The panel is synced to the time slider — moving the slider updates the highlighted log entry, and clicking a log entry seeks the slider.
- **Motion planner naming** (`planners.py`, `pipette.py`): each `PipetteMotionPlanner` subclass now provides a `_default_name()` so that `move_requested` events carry a meaningful name (e.g. `"move to target"`, `"search"`, `"move to home"`). The planner name is also forwarded through `_movePath` so the `sigMoveRequested` signal carries it.
- **`_autoFindTip` refactor** (`multipatch.py`): switched from a single `findTipInFrame` + `setTipOffsetIfAcceptable` call to `pip.iterativelyFindTip(max_reps=10)`, enabling multi-iteration auto-tip-finding from the MultiPatch module UI.

## Test plan

- [x] Open the MultiPatch module and confirm an event log appears inside each pipette row beneath the clamp controls
- [x] Change pipette states and confirm events (state changes, moves, calibration) appear in the per-pipette log in real time with correct relative timestamps
- [x] Click "Reset History" and confirm all per-pipette event logs clear and restart their timers
- [ ] Verify the splitter between the left controls panel, pipette matrix, and plots panel is draggable and that its position is saved and restored across module close/reopen
- [x] Open a saved MultiPatch log file and confirm the event log panel appears below the plots, is populated correctly from all devices, and is sorted by time
- [ ] Move the time slider in the log viewer and confirm the highlighted log entry tracks it; click a log entry and confirm the time slider jumps to that time
- [x] Trigger "Auto-find tip" and confirm it iterates multiple frames rather than checking only once
- [x] Save tip image button state should now correctly reflect the saved config value on reload (not always check itself on)
- [x] Confirm that "Set tip" and "Set target" via camera click still work correctly with array-indexed scope position

## Dependencies

> - #506 — `iterativelyFindTip()` is defined there (called in `_autoFindTip`)


outofculture (commented):
A few issues found in this diff:

**1. `patchpipette.py` — `_logEvents` list mutated from multiple threads without a lock**

`emitNewEvent()` appends to `self._logEvents` and may be called from background threads (test pulse callbacks, state machine threads). `newPatchAttempt()` calls `.clear()`. No lock protects these operations. A `.clear()` racing with an `.append()` can corrupt the list. `eventLog()` makes a copy (correct), but the mutation paths are unprotected.

**2. `pipetteEventLog.py` `format_event()` — bare `['info']` dict access on `state_event`**

The `state_event` branch does `return event_data['info']`. Every other branch uses `.get()` with a fallback. If any `state_event` record is missing `'info'` (corrupted log file, old format, different emitter), this raises `KeyError` and crashes the UI update path.

**3. `MultiPatchLog.py` — event log likely always empty in replay mode**

The `'log_events'` slot in `MultiPatchLogData` is initialized as `list(range(N))` but the data pipeline doesn't appear to call `_prepare_event_for_use` for `log_events`. The `_populateEventLog` guard `isinstance(entry, tuple) and len(entry) == 3` will never match the placeholder integers, so the event log panel is always empty when replaying a log file.

**4. `multipatch.py` + `pipetteControl.py` — Reset History clears UI but not device buffer**

`resetHistory()` calls `ctrl.clearEventLog()` (resets the UI start time) but does NOT call `pip.newPatchAttempt()` (which clears `self._logEvents`). After Reset History, if the module is closed and reopened, `PipetteControl.__init__` re-reads `pipette.eventLog()` and repopulates from the full uncleared buffer. The UI and device log diverge.

**5. `planners.py` — `SearchMotionPlanner._default_name()` is dead code**

`SearchMotionPlanner._move()` hardcodes the name string in two places. The `self.name` attribute set via `_default_name()` is never used in this class, so the planner-naming refactor has no effect for search moves.

PR #508, opened 2026-04-01
feat: add WebcamCamera device driver for USB webcams via OpenCV

outofculture:
## Summary

- Adds a new `WebcamCamera` device driver (`acq4/devices/WebcamCamera/`) that integrates any OpenCV-accessible camera as a first-class acq4 camera device
- Camera selection supports either a numeric device index (`cameraIndex`) or a substring match on the device description (`cameraName`); on Windows, human-readable names are resolved via `pygrabber` (DirectShow), with a generic index-probe fallback for other platforms
- Frame acquisition runs on a dedicated background thread that continuously reads from `cv2.VideoCapture` and pushes frames into a bounded queue; the capture object is released on `stopCamera()` and re-opened on `startCamera()`
- Output frames can be returned as grayscale or BGR (`colorMode` config option); frames are transposed to match acq4's (x, y) convention
- Exposes a fixed set of standard camera parameters (`exposure`, `triggerMode`, `binning`, `region`, `sensorSize`, `bitDepth`) plus a dynamic set of optional OpenCV properties (brightness, contrast, gain, gamma, focus, auto-exposure, white balance, pixel format/FOURCC, frame rate, etc.) that are registered only if the driver reports them as available and readable
- ROI and binning are fixed to full-frame / 1×1 and silently corrected when `autoCorrect=True`, matching how other acq4 camera drivers handle unsupported features
- Configurable options include: `backend` (OpenCV capture backend), `resolution`, `colorMode`, `bufferSize`, `pixelFormat` (initial FOURCC), and `pixelFormats` (list of allowed FOURCC values exposed in the UI)

## Test plan

- [ ] Configure a USB webcam with `cameraIndex` in `devices.cfg` and confirm it appears in the Camera module
- [ ] Configure a webcam using `cameraName` (substring match) and confirm correct device is selected
- [ ] Acquire live frames and verify they display correctly in the camera module viewport
- [ ] Confirm `stopCamera()` releases the capture resource and `startCamera()` re-opens it
- [ ] Set `colorMode: bgr` and verify color frames are returned; set `colorMode: gray` and verify grayscale
- [ ] Attempt to set `region` or `binning` to non-default values with `autoCorrect=True` and confirm they are silently reset
- [ ] Verify optional cv2 parameters (e.g. brightness, gain) appear in the UI only when the driver reports them as available
- [ ] Test on Windows with `pygrabber` installed and confirm human-readable device names are used for `cameraName` matching
- [ ] Omit both `cameraIndex` and `cameraName` from config and confirm an informative error listing available devices is raised

### DIFF
--- a/acq4/devices/WebcamCamera/webcam_camera.py
+++ b/acq4/devices/WebcamCamera/webcam_camera.py
@@ -505,7 +505,12 @@ class WebcamCamera(Camera):
                 self._capture = None
 
     def newFrames(self):
-        frames = [self.frameQueue.get() for _ in range(self.frameQueue.qsize())]
+        frames = []
+        while True:
+            try:
+                frames.append(self.frameQueue.get_nowait())
+            except queue.Empty:
+                break
         return frames
 
     def _acquisitionLoop(self):
--- a/acq4/devices/WebcamCamera/webcam_camera.py
+++ b/acq4/devices/WebcamCamera/webcam_camera.py
@@ -1,0 +1,570 @@
+from __future__ import annotations
+
+from collections import OrderedDict
+import queue
+import threading
+import time
+
+import numpy as np
+
+import acq4.util.ptime as ptime
+from acq4.devices.Camera import Camera
+
+try:
+    import cv2
+except ImportError:
+    cv2 = None
+
+
+class WebcamCamera(Camera):
+    """
+    Minimal camera driver using OpenCV VideoCapture.
+
+    This implementation intentionally keeps behavior simple:
+    - Uses a configured capture resolution as the effective sensor size.
+    - Streams frames continuously via ``newFrames``.
+    - Keeps trigger mode, ROI, and binning fixed to basic values.
+    - Exposes a small set of optional cv2 controls when available.
+
+    Configuration options (exactly one of ``cameraIndex`` or ``cameraName`` must be provided):
+
+    - ``cameraIndex`` (int): Webcam index for cv2.VideoCapture.
+    - ``cameraName`` (str): Substring of the camera's device description (case-insensitive).
+      If omitted along with ``cameraIndex``, an exception is raised that lists all
+      available ``index: description`` pairs so you can copy one into the config.
+    - ``backend`` (int | str, optional): OpenCV backend (for example "CAP_DSHOW").
+    - ``resolution`` ([w, h], default [640, 480]): Requested capture size used as effective sensor size.
+    - ``colorMode`` ("gray" | "bgr", default "gray"): Output frame format.
+    - ``bufferSize`` (int, optional): Requested driver buffer size.
+    - ``pixelFormat`` (str, optional): Initial FOURCC string (for example "MJPG").
+    - ``pixelFormats`` ([str], optional): Allowed pixel format values exposed in UI.
+    """
+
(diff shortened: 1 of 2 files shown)


outofculture (commented):
## Code Review: `webcam_camera.py`

Several issues in the new WebcamCamera driver that should be addressed before merge:

**1. `newFrames()` — race condition on `qsize()` + potential hang**

`frames = [self.frameQueue.get() for _ in range(self.frameQueue.qsize())]`

`qsize()` is not atomic with `get()`. If the background thread enqueues items between reading `qsize()` and consuming them, the list comprehension blocks indefinitely when the camera is stopped. Use `get_nowait()` in a loop with a `queue.Empty` catch instead.

**2. `_acquisitionLoop()` — `camLock` held across blocking `cap.read()`**

`self._capture.read()` can block for a full frame period (33 ms+ at 30 fps) while `camLock` is held. `stopCamera()`, `startCamera()`, and all `_writeCv2Param()`/`_readCv2Param()` calls also acquire `camLock`, so any parameter change or stop request will stall for a full frame duration. The standard pattern (see `VimbaXCamera`) is to not hold the lock across the frame read itself.

**3. `_cvAcqThread` — no stop mechanism, resource leak**

`_acquisitionLoop` is `while True:` with no stop flag, event, or condition. The thread runs until process exit — there's no clean shutdown path. If the device is re-instantiated (e.g., in tests), the old thread keeps running against a released capture object. A `threading.Event` stop flag should be checked in the loop.

**4. `frameQueue` — blocking `put()` stalls acquisition thread**

`self.frameQueue.put(out)` with no timeout blocks indefinitely when the queue is full (`maxsize=100`). If `newFrames()` is not called promptly, the acquisition thread stalls and holds the lock on the next iteration. Use `put_nowait()` with a drop-and-warn, or `put(timeout=...)`.

**5. `cap.read()` runs even when camera is logically stopped**

`_acquisitionLoop` reads frames continuously regardless of whether `startCamera()` has been called. Frames are produced and queued even when the `AcquireThread` hasn't started acquisition, bypassing the normal camera gating.

**6. `exposure` parameter stored but never applied to camera**

Setting `exposure` stores the value and reports it back, but never calls `cap.set(cv2.CAP_PROP_EXPOSURE, ...)`. The `AcquireThread` uses the stored `exposure` value to calculate its no-frame timeout (`10 + exposure`). With a real exposure of e.g. 5 s but stored value of 0.03 s, `AcquireThread` will incorrectly time out after ~10 seconds.

**7. `_acquireFrames()` — dead code**

This method is defined but never called anywhere. It duplicates the frame-reading logic already in `_acquisitionLoop` and will silently diverge if the frame format changes. Either wire it in or remove it.

acq4/devices/WebcamCamera/webcam_camera.py
```diff
@@ -0,0 +1,577 @@
...
+            try:
+                with self.camLock:
+                    if self._capture is None:
+                        ok = False
+                        frame = None
+                    else:
+                        ok, frame = self._capture.read()
+                now = ptime.time()
+                if not ok or frame is None:
+                    time.sleep(0.01)
```
outofculture, line 527: 10ms seems a little fast for an idle loop. can it decay down to 1s after not being used for a while?

PR #509, opened 2026-04-01
feat: add StreamDock device driver for VSDInside Stream Deck hardware

outofculture:
## Summary

- **New `StreamDock` device** (`acq4/devices/StreamDock/`): integrates with VSDInside Stream Dock hardware via the Stream Deck plugin WebSocket protocol. The device runs a background thread to maintain the WebSocket connection to the VSD Craft software, dispatches button events via a `sigButtonEvent` Qt signal, and exposes `setTitle`/`setImage` methods for updating key displays.
- **Bridge plugin** (`bridge.py` + `manifest.json`): a minimal Python script that is automatically installed into the VSD Craft plugins directory on first startup. It writes a handoff file (`~/.acq4_streamdock_port.json`) so ACQ4 can discover the WebSocket port, then holds the connection open so VSD Craft doesn't consider the plugin crashed.
- **Plugin installer** (`plugin.py`): handles copying `bridge.py`, `manifest.json`, and placeholder PNG icons into the VSD Craft plugins folder, and optionally launching VSD Craft if it isn't already running.
- **`stream_dock` interface protocol**: devices can declare themselves as `stream_dock` implementers via `man.declareInterface(name, ['stream_dock'], self)` and implement `configure_dock(stream_dock_device)`. `StreamDock` auto-discovers these at startup and calls `configure_dock` on each.
- **`DoverStage` integration**: registers X/Y/Z axis enable/disable toggle buttons with the Stream Dock. Button backgrounds are green when an axis is enabled and red when disabled. PNG images are generated programmatically using only stdlib (`struct`, `zlib`, `base64`) — no Pillow required.
- **`DoverStage` declares `stream_dock` interface**: so `StreamDock` picks it up automatically without manual wiring in config.
- **`smartstage.py` minor fix**: adds an optional `name` parameter to `SmartStage.move()` for diagnostic log messages; also fixes a trailing-whitespace style issue in the docstring.
- **`Sonicator.py` temporary debug change**: `safeToSonicate` is short-circuited to always return `True` with a TODO comment, and a debug `print` statement was added. These appear to be in-progress debugging changes, not intentional production behaviour.

## Test plan

- [ ] Configure a `StreamDock` device in `devices.cfg` and start ACQ4 — verify the bridge plugin is auto-installed into the VSD Craft plugins directory
- [ ] Restart VSD Craft after first install; confirm the ACQ4 plugin appears and ACQ4 connects (check logs for "Handoff received")
- [ ] Add `DoverStage` to the config alongside `StreamDock`; confirm X/Y/Z toggle buttons appear on the hardware with green backgrounds
- [ ] Press an axis button and verify the axis is toggled (enabled → disabled or vice versa) and the button colour updates accordingly
- [ ] Verify that `SmartStage.move()` still works correctly with and without the new `name` keyword argument
- [ ] Confirm that the `Sonicator.py` debug short-circuit (`return True`) is intentional or is removed before merge

## Dependencies

> - #513 — fills in the `reserver=` argument for all `device.reserve()` call sites that #510 makes required


outofculture (commented):
Thanks for the PR! A few issues to address before merging:

1. **`streamdock.py` — 6 raw `print()` statements in production code**
   In `_on_open`, `_on_message`, `_on_error`, `_on_close`, and `setTitle`, there are `print()` calls that duplicate or precede `self.logger.*` calls. These are debug-era leftovers that will pollute stdout in production. The logger calls are sufficient; the `print()` calls should be removed.

2. **`plugin.py` / `streamdock.py` — `RuntimeError` on non-Windows when `autoInstall=True` (the default)**
   `auto_install_plugin()` checks for VSD Craft at a hardcoded Windows path (`C:/Program Files (x86)/VSD Craft`) and raises `RuntimeError` if not found. Since `autoInstall` defaults to `True`, `StreamDock.__init__()` will raise on any Linux/macOS system. Either default `autoInstall` to `False`, or skip silently on non-Windows rather than raising.

3. **`plugin.py` — `tasklist`/`wmic` called unconditionally, crash on non-Windows**
   `vsdcraft_is_running()` uses `tasklist` and `is_bridge_running()` uses `wmic` — both Windows-only CLI tools. They are called from the `StreamDock-handoff` background thread. On Linux/macOS, `subprocess.run(['tasklist', ...])` raises `FileNotFoundError`, silently killing the thread. The device then hangs indefinitely waiting for a handoff that never arrives.

4. **`doverstage.py` — button color callback wired to wrong method name**
   The commented-out line `# self.dev.set_enabled_callback(_update_colors)` uses `set_enabled_callback`, but the actual method on `SmartStage` is `add_enabled_state_callback()`. The `_update_colors` closure is defined but never registered, so button colors never update when axis enable state changes externally. Either fix the method name and uncomment, or remove the dead closure.

5. **`doverstage.py` — unused `import numpy as np`**
   `numpy` is imported but not referenced anywhere in the file after this PR. This is a dead import.

acq4/devices/Sonicator.py
```diff
@@ -37,6 +37,7 @@ def __init__(self, deviceManager, config, name):
         self.protocols = config.get("protocols", {})
 
     def safeToSonicate(self, _future: Future = None, askUser=True) -> bool:
+        return True  # TODO unbroke this
```
outofculture: not really part of this PR

PR #510, opened 2026-04-01
fix: named device reservation with deadlock diagnostics, imaging record_thread, and misc device fixes

outofculture:
## Summary

- **Device.py — named reservation with deadlock diagnostics**: The device lock API now requires a `reserver` name on every `reserve()` / `reserved()` call. The old single traceback `_lock_tb_` is replaced by a stack of `(reserver_name, traceback)` entries, so timeout errors now clearly identify who holds the lock. A new thread-local `thread_device_locks` registry tracks which devices each thread currently holds; acquiring a new device while already holding others emits a warning to catch potential deadlocks early.
- **Device.py — new `DeviceLocker` class**: Adds a context-manager helper that locks multiple devices in sorted name order (the standard deadlock-prevention strategy), with the same deadlock-warning logic. Intended to replace ad-hoc multi-device locking patterns.
- **DAQSonicator.py — dynamic sample rate**: Replaces the hard-coded 1 MHz sample rate with a live query to the DAQ hardware (`GetDevAOMaxRate`). Also passes `reserver="DAQSonicator"` to `reserveDevices()` to take advantage of the new named-reservation API. Extracts `_protocolDuration()` as a standalone helper so `Sonicator.doProtocol` can apply a timeout.
- **Sonicator.py — protocol timeout and `_protocolDuration` abstract method**: `doProtocol` now passes `timeout=duration * 1.3` to `waitFor`, preventing indefinite hangs. Adds `_protocolDuration()` as an abstract method that subclasses must implement.
- **Sonicator.py — temporary safety bypass**: `safeToSonicate` is short-circuited to `return True` with a TODO comment while the containment-point check is being debugged; a `print` statement is left in for that same debugging.
- **LEDLightSource — config documentation update**: Expands and corrects the example configuration docstring to show the two-device pattern (a `DAQGeneric` channel device plus the `LEDLightSource` device), updated channel references, and clearer annotations.
- **imaging/record_thread.py — transform handling fix**: Stops calling `getImage()` before storing frames in the collection list (defers it to write time). Uses `SRT3DTransform.from_affine` to extract translation from arbitrary affine transforms rather than calling `.getTranslation()` directly (which was fragile for non-SRT transforms). Uses `frame.serializableInfo()` when writing the first file in a stack.

## Test plan

- [ ] Verify that `Device.reserve()` raises `ValueError` when called without a `reserver` argument
- [ ] Verify that lock timeout errors now include the reserver name and traceback in the exception message
- [ ] Verify that acquiring a second device while holding one produces a deadlock warning in the log
- [ ] Verify that `DeviceLocker` locks multiple devices in alphabetical order and releases them all on exit
- [ ] Run a DAQSonicator protocol and confirm the sample rate matches the hardware's reported AO max rate
- [ ] Run a Sonicator protocol and confirm it times out rather than hanging if the hardware stalls
- [ ] Record an imaging sequence and confirm frames are written correctly (no dropped frames, correct translations in metadata)
- [ ] Confirm LEDLightSource initialises correctly with the two-device config pattern shown in the updated docstring

## Dependencies

> - #513 must also be merged to fix call sites in `Manager.py`, `nidaq.py`, and others that do not yet pass `reserver=`


outofculture (commented):
## Code Review Notes

### 1. `Device.py` — `thread_device_locks` global dict leaks thread entries; stale state inherited

Thread IDs from `threading.get_ident()` are reused by the OS after a thread exits. Entries in `thread_device_locks` are never cleaned up when a thread dies. A new thread that happens to reuse a dead thread's ID inherits a stale `{device_name: lock_count}` mapping, which can cause the deadlock warning to fire spuriously — or, worse, suppress a real one. Recommend switching to `threading.local()` instead of a global dict keyed by thread ID.

### 2. `Device.py` — `DeviceLocker.tryLock()` graceful-failure path is unreachable

When `block=True`, `Device.reserve()` either returns `True` or raises `TimeoutError` — it never returns `False`. The `if not devLocked: ... return False` branch is dead code. A timeout will propagate as an unhandled exception instead of the clean `lockErr` message the code intends. Either catch `TimeoutError` explicitly and return `False` there, or document that the `block=True` path always raises on failure.

### 3. `Device.py` — `release()` leaves diagnostic state desync'd if `unlock()` throws

The bare `except` in `release()` swallows any exception from `self._lock_.unlock()`, but the `_lock_stack_.pop()` and thread-counter decrement that follow are skipped when an exception is raised. After enough reserve/release cycles this causes the lock stack to grow without bound, and the timeout error message will show the wrong reserver name. The cleanup of `_lock_stack_` and the thread counter should happen in a `finally` block (or before the potentially-throwing call) so diagnostic state stays consistent regardless of whether `unlock()` succeeds.

### 4. `DAQSonicator.py` — protocol parsed twice via `load_stimulus`

`doProtocol` calls `_protocolDuration(protocol)` and then `_doProtocol(protocol)`, and both functions call `load_stimulus(protocol)` internally. If `load_stimulus` is stateful or mutates its argument, `_doProtocol` may receive a corrupted or already-consumed protocol. At minimum this is wasteful. Worth clarifying `load_stimulus`'s contract — if it is pure and idempotent the redundancy is merely inefficient; if it mutates state the second call may silently misbehave.

### 5. `DAQSonicator.py` — channel string split to extract device name is fragile

```python
daq_dev_name = self._daq.config["channels"]["command"]["channel"].split("/")[1]
```

This assumes the channel string starts with a leading slash (e.g. `"/Dev1/ao0"`). If the string is in the more common NI-DAQmx form `"Dev1/ao0"` (no leading slash), `split("/")[1]` returns `"ao0"` instead of `"Dev1"`, silently passing the wrong name to `GetDevAOMaxRate` with no validation afterward. Consider using `lstrip("/").split("/")[0]` or splitting on `/` and taking the first non-empty part, and add an assertion or log warning if the result looks implausible.

---

PR #500, opened 2026-04-01
4-axis Sensapex IK, stage hysteresis correction, geometry utilities

outofculture:
## Summary
- Adds inverse kinematics support for 4-axis Sensapex manipulators (diagonal axis decomposition)
- Adds configurable hysteresis correction for focus/z-axis moves in the imaging sequencer (`hysteresisCorrection` device property)
- Adds geometry utilities (`Plane`, `Geometry`, `load_transform_from_anything`) used by IK and future spatial features
- Includes test coverage for 4-axis stage behavior

## Test plan
- [x] Run `acq4/devices/Stage/tests/test_4_axis_stage.py`
- [ ] Verify hysteresis correction engages/disengages correctly per device config
- [x] Manual test with physical 4-axis Sensapex manipulator if available

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #511, opened 2026-04-01
refactor: migrate OptomechDevice transforms to coorx and add adapter-based Visualize3D integration

outofculture:
## Summary

- **Transform system migration**: `OptomechDevice` now uses `coorx` transforms (`SRT3DTransform`, `AffineTransform`) throughout, replacing `pg.SRTTransform3D`. Inverse and global transforms are now computed via `.inverse` and chained multiplication rather than manual cache entries. The `_mapTransform` helper is replaced by a module-level `map_through_transform` function. Cached inverse transforms (`__inverseTransform`, `__inverseGlobalTransform`, etc.) are removed; only `__globalTransform` and `__globalPhysicalTransform` are cached.
- **Visualize3D adapter pattern**: Devices no longer register themselves directly with the `Visualize3D` module. Instead, `OptomechDevice` now declares the `"3D Visualizable"` interface and implements `visualize3DAdapter(win)`, which returns an `OptomechDeviceVisualizerAdapter`. `Visualize3D` discovers devices via interface list changes and calls `visualize3DAdapter()` for each.
- **`VisualizerWindow` GUI overhaul**: The right-hand panel is now a `pyqtgraph.ParameterTree` instead of a `QTreeWidget`. Device management methods (`addDevice`, `_removeDevice`, `handleTransformUpdate`, etc.) are replaced by `addAdapter`, `removeAdapter`, and `findAdapter(predicate)`. `OptomechDeviceVisualizerAdapter` manages its own mesh, bounds visualization, and parameter tree entries.
- **`PipetteVisualizerAdapter`** added in new `acq4/devices/Pipette/visualization.py`: extends `OptomechDeviceVisualizerAdapter` with pipette-specific visualization — planned path, actual motion path, target marker, error highlighting, and a history slider for reviewing past path plans.
- **`MovePathException`** introduced in `travelers_proxy.py`: raised by the path executor when a move fails, carries the path and failing point so the pipette can offset the points and hand them to the visualizer for error display. `VisualizePathPlan` is renamed `VisualizePathSearch`.
- **`Manager.py` cleanups**: Logging setup refactored — `setup_logging(TEMP_LOG)` call and temp-log migration logic moved into `set_log_file()`; `--log-level`/`--root-log-level` CLI args removed. `DeviceLocker` moved to `Device.py`. `getOrLoadModule(name)` added for lazy module loading. `reserveDevices()` gains a `reserver` parameter for deadlock diagnostics. Window shortcut now un-minimizes the window before raising it. Shutdown log messages promoted from DEBUG to INFO.
- **Pipette additions and fixes**: `iterativelyFindTip()` method added for convergent tip-finding loops. `resetGlobalPosition()` corrected to compute offset in manipulator coordinates. `_updateTransform()` uses `AffineTransform.set_mapping()`. `focusTip`/`focusTarget` use `imagingDevice()` instead of `scopeDevice()`. `saveManualTipPosition` now saves axis yaw and pitch alongside tip position, and uses a timestamped filename. `getRecordingChambers`/`getCleaningWell` return type updated to `InteractionSite`. `getNucleusDepositionWell()` and `getElectrodeSolutionWell()` added. Motion planners gain `_default_name()` overrides for clearer move names.
- **`planners.py`**: Path validation now raises `ValueError` (instead of asserting) for non-finite positions; `checkGlobalLimits` now receives the `linear` flag. Error paths are visualized via the adapter. `getOrLoadModule("Visualize3D")` replaces `getModule(...)`.

## Test plan

- [ ] Open `Visualize3D` module and confirm that all `OptomechDevice` instances appear in the `ParameterTree` with working visibility toggles
- [ ] Toggle `Geometry`, `Range of Motion`, and per-device visibility on/off for at least two devices
- [ ] Click `Center View` for a device and verify the 3D camera moves to that device's position
- [ ] Run a path plan for a pipette and verify the planned path and actual motion path appear in the 3D view
- [ ] Trigger a path planning failure (e.g., move to an out-of-range position) and verify the error path is highlighted and the window comes to front
- [ ] Use the `History` slider to review previous path plans
- [ ] Confirm `Visualize3D` can be opened after devices are loaded (lazy via `getOrLoadModule`)
- [ ] Confirm no errors when `Visualize3D` is closed and then a pipette move is attempted
- [ ] Verify `iterativelyFindTip` converges and raises `TimeoutError` when it cannot
- [ ] Confirm `saveManualTipPosition` saves yaw/pitch alongside tip position and uses timestamped filename
- [ ] Verify window keyboard shortcut un-minimizes the window before raising it
- [ ] Run existing unit tests for `OptomechDevice` transforms to confirm coorx-backed transforms behave correctly

## Dependencies

> - #500 (axis-of-diagonevil) — extends `Stage.checkGlobalLimits()` to accept the `linear` argument that `planners.py` now passes
> - #501 (step-into-my-chambers) — renames RecordingChamber to InteractionSite

### DIFF
--- a/acq4/Manager.py
+++ b/acq4/Manager.py
@@ -24,15 +24,13 @@ from pyqtgraph.util.mutex import Mutex
 from . import __version__
 from . import devices, modules
 from .Interfaces import InterfaceDirectory
-from .devices.Device import Device, DeviceTask
-from .logging_config import get_logger, setup_logging, HistoricLogRecord
+from .devices.Device import Device, DeviceTask, DeviceLocker
+from .logging_config import get_logger, set_log_file, setup_logging, HistoricLogRecord
 from .util import DataManager, ptime, Qt
 from .util.DataManager import DirHandle
 from .util.HelpfulException import HelpfulException
 from .util.LogWindow import get_log_window, get_error_dialog
 
-TEMP_LOG = "temp_log.json"
-setup_logging(TEMP_LOG, gui=False, console_level=logging.DEBUG)
 logger = get_logger()
 
 
@@ -71,8 +69,6 @@ class Manager(Qt.QObject):
         parser.add_argument("--module", "-m", help="Module name to load", action="append")
         parser.add_argument("--base-dir", "-b", help="Base directory to use")
         parser.add_argument("--storage-dir", "-s", help="Storage directory to use")
-        parser.add_argument("--log-level", action="store", help="Set the console log level", default="WARNING")
-        parser.add_argument("--root-log-level", action="store", help="Set the root log level", default="DEBUG")
         parser.add_argument("--disable", "-d", help="Disable the device specified", action="append")
         parser.add_argument("--disable-all", "-D", help="Disable all devices", action="store_true")
         parser.add_argument("--exit-on-error", "-x", help="Whether to exit immidiately on the first exception during initial Manager setup", action="store_true")
@@ -105,8 +101,6 @@ class Manager(Qt.QObject):
         self.taskLock = Mutex(Qt.QMutex.Recursive)
         self._folderTypes = None
         self._logFile = None
-        self._consoleLogLevel = logging.WARNING
-        self._rootLogLevel = logging.DEBUG
 
         try:
             if Manager.CREATED:
@@ -141,8 +135,6 @@ class Manager(Qt.QObject):
         self.exitOnError = args.exit_on_error
         self.disableDevs = args.disable or []
         self.disableAllDevs = args.disable_all
-        self._consoleLogLevel = getattr(logging, args.log_level.upper(), logging.WARNING)
-        self._rootLogLevel = getattr(logging, args.root_log_level.upper(), logging.DEBUG)
 
         self.configDir = os.path.dirname(args.config)
         self.readConfig(args.config)
@@ -170,9 +162,6 @@ class Manager(Qt.QObject):
                         # we have to show it now, otherwise we'll have no windows
                         self.showGUI()
                     raise
-            setup_logging(
-                TEMP_LOG, acq4_level=self._rootLogLevel, console_level=self._consoleLogLevel
-            )
 
         except Exception:
             if self.exitOnError:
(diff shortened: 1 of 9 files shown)


outofculture (commented):
The coorx migration and Visualize3D adapter work looks like a good structural direction. A few issues worth resolving before merge:

**1. `planners.py` — `checkGlobalLimits()` called with wrong arity (hard blocker)**

The PR passes a second argument `linear` to `self.manipulator.checkGlobalLimits(manipulatorGlobalPos, linear)`, but `Stage.checkGlobalLimits` in `Stage.py` only accepts one positional argument. This will raise `TypeError` on every path-validation attempt, breaking all manipulator movement. This signature change is in PR #500 (axis-of-diagonevil) — add that as a dependency.

**2. `Manager.py` `reserveDevices()` — string device names no longer resolved**

The string-to-device resolution (`self.getDevice(d) if isinstance(d, str) else d`) was removed. The docstring example still shows string names being passed. Any existing caller using string names will fail when `DeviceLocker` tries to call `.name()` on a bare string. Also confirm `DeviceLocker` is properly exported from `Device.py`.

**3. `pipette.py` `_moveToGlobal()` — `getOrLoadModule` called on every motion**

`self.dm.getOrLoadModule("Visualize3D")` is called on every single pipette move. If Visualize3D isn't running, this may attempt to load it (I/O, potential blocking) inside a motion-critical code path. Use a non-loading existence check first.

**4. `OptomechDevice.py` `invalidateCachedTransforms()` — `invalidateLocal` parameter is a silent no-op**

The parameter remains in the signature but the body no longer references it. A parameter that appears meaningful but does nothing is confusing for callers. Either remove it or document explicitly that it's a no-op.

**5. `pipette.py` `saveManualTipPosition()` — `is_below_surface` ternary always evaluates identically**

`scan_dist = (2 + (np.random.random()**2) * (100 if is_below_surface else 100)) * 1e-6` — both branches are `100`. The `is_below_surface` check has no effect. Looks like a typo.

**6. `pipette.py` `iterativelyFindTip()` — calibration saved on cancellation, not just on error**

The bare `except Exception` catch calls `self.setTipOffset(start_pos)` (writes calibration to disk) on any exception including user stop requests. Use a narrower exception type or check for stop-request exceptions before saving.

**7. `travelers_proxy.py` — TODO comment explicitly says the code will break**

An active code path with `# TODO this is going to break with bitrot` is a merge blocker. Fix it or remove the code path.

**8. `OptomechDevice.py` — module-level import of UI module from device layer**

`from acq4.modules.Visualize3D import Visualize3D` is only used to read a string constant (`interfaceName`). Replace with a plain string literal to remove the devices→modules dependency inversion.

**9. `planners.py` — raw `print()` in production code**

`print(f"cache primed for {self.pip.name()}")` should use the logger.

outofculture:
this pr is half-baked...

PR #512, opened 2026-04-01
refactor: extract AutomationDebug logic into focused helper classes

outofculture:
## Summary

- Extracts the large monolithic `AutomationDebugWindow` class into four focused helper classes: `Autopatcher`, `CellDetector`, `FeatureTracker`, and `MockDataHandler`
- `AutomationDebug.py` shrinks from ~1000 lines to ~100 lines; the main window now wires UI events to the composed helpers rather than implementing logic itself
- `autopatch.py` (`Autopatcher`): the autopatch demo loop gains several improvements — creates a per-run data directory, integrates with the MultiPatch module for recording, uses `iterativelyFindTip()` instead of the old three-call tip-finding sequence, and saves GFP/brightfield image stacks after whole-cell success
- `detection.py` (`CellDetector`): cell detection, bounding-box display, mock/real z-stack acquisition, and ranking window logic extracted from the main window
- `feature_tracking.py` (`FeatureTracker`): pipette calibration test and feature tracking extracted from the main window
- `mock_data.py` (`MockDataHandler`): MetaArray mock-file loading extracted from the main window
- `window.ui`: adds configurable z-stack start/stop depth spinboxes to the Cell Detection group box
- Removes the empty `cell.py` file and the now-unused inline `RankingWindow` class definition from the main module file
- Fixes `_setTopLeft` / `_setBottomRight` to use tuple-based coordinate mapping instead of `QPointF` (compatible with 3D transforms)

## Test plan

- [ ] Launch the `AutomationDebug` module and verify the UI loads without errors
- [ ] Run z-stack cell detection (real and mock) and confirm bounding boxes appear
- [ ] Run the autopatch demo and verify it creates a data directory, records via MultiPatch, and saves image stacks after whole-cell success
- [ ] Test the "Track Features" button and pipette calibration test button
- [ ] Confirm the z-stack depth spinboxes are visible in the Cell Detection group and their values are respected during acquisition
- [ ] Verify the "Set Top Left" and "Set Bottom Right" buttons set the correct coordinates

## Dependencies

> - #503 — defines `Microscope.moveDip()` (called in `_autopatchDemo`)
> - #506 — defines `iterativelyFindTip()` and the `startANewCell=` kwarg for `setState()` (both called in `autopatch.py`)


outofculture (commented):
A few issues worth addressing before merge:

**1. `mock_data.py` `_create_mock_stack_from_file()` — possible in-place mutation of source transform**

`mock_frame_transform = live_frame_global_transform.full_matrix` followed by `mock_frame_transform[2, 2] = step_z` — if `.full_matrix` returns a numpy view, this silently corrupts `live_frame_global_transform` on the first loop iteration. The previous code used `pg.SRTTransform3D(live_frame_global_transform.saveState())`, which was an explicit copy. Add `.copy()`.

**2. `autopatch.py` `_autopatchFindCell()` — dead code after `return None`, autonomous detection broken**

Several lines after the early `return None` were intended to trigger autonomous cell detection when the cell list is exhausted. They're unreachable, so `_autopatchDemo` immediately logs "No cells found; quitting demo" instead of initiating detection. This is a behavioral regression from the original.

**3. `autopatch.py` — `logger.warning` used for routine informational events**

"Autopatch: Start cell patching", "Autopatch: Cell patching finished", "Autopatch: Next cell!", and "Autopatch: Task runner sequence completed" are state transitions, not warnings. These should be `logger.info`.

**4. `detection.py` `_rankingWindowClosed()` — `logger.exception` with no active exception**

`logger.exception(...)` is called inside an `except ValueError` block after the error has been handled, with no active traceback in scope. This emits a null/misleading traceback. Should be `logger.warning(...)`.

**5. `autopatch.py` — large commented-out TODO block should not be committed**

A ~20-line block of commented-out nucleus collection/resealing code with `# TODO too slow for today's demo` and `# TODO once we have motion planning` should not be committed. Either implement it or remove it.

PR #535, opened 2026-06-02
feat: save cell-health detection sessions (dir config + z-stack + cellpose masks)

outofculture:
## Summary

Persist the full output of the AutomationDebug **"Detect Neurons"** workflow so its annotation sessions are organized and reloadable for training. Two parts:

### 1. Configurable save directory
- New `misc/cellAnnotationDir` config option, resolved in `detection.py` via `_annotation_save_dir()`: configured value wins, else a `cell_annotations` subdirectory of the data storage directory, else `None` (tool falls back to the working directory).
- Documented in `config/example/default.cfg`.

### 2. Save the z-stack and cellpose masks
Previously the in-memory detection stack and its cellpose segmentation were discarded, so sessions couldn't be reloaded or retrained. Now `detection.py` picks one timestamped base name per session and:
- passes `save_prefix` to `detect_neurons`, which writes `{base}.ma` (the z-stack as it arrived) and `{base}_cellpose_masks.npy` (the masks, immediately after creation, in the detection process — no large-array round-trip);
- reuses the same `save_dir`+`base_name` when opening the annotation tool, which writes `{base}_annotations.json` referencing those files.

The three files share the annotation tool's naming scheme, so a session reloads as a normal file-based session. Orientation is safe: `do_neuron_detection`'s `data` is the same array the tool displays, so `center_ijk` indexes the saved `.ma` and the masks align.

## Cross-repo dependency

The matching `acq4-automation` change (branch `save-stack-and-masks`: `detect_neurons` `save_prefix`, `CellAnnotationTool` `base_name`) must be merged first — without it the `save_prefix` argument doesn't exist.

## Testing

`detection.py` compiles; `_annotation_save_dir` resolution verified in isolation. The acq4-automation side has unit tests for the `.ma`/masks round-trip and the tool's reload wiring (29 passing). The GUI launch path needs a running Manager + camera, so it isn't covered by an automated test here.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/acq4/modules/AutomationDebug/detection.py
+++ b/acq4/modules/AutomationDebug/detection.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import datetime
 from pathlib import Path
 from typing import TYPE_CHECKING
 
@@ -164,6 +165,20 @@ class CellDetector:
 
         win.cameraDevice.setFocusDepth(depth, name=f"{win.cameraDevice.name()} restore focus after detection z-stack")  # Restore focus
 
+        # Persist this detection session under one base name so the saved z-stack,
+        # cellpose masks, and annotations share the annotation tool's naming scheme
+        # and reload together. The base is shared with the annotation tool launched
+        # in _handleDetectResults.
+        win._annotation_save_dir = _annotation_save_dir(man)
+        win._annotation_base_name = datetime.datetime.now().strftime(
+            "in_memory_stack_%Y%m%d_%H%M%S_%f"
+        )
+        save_prefix = (
+            str(win._annotation_save_dir / win._annotation_base_name)
+            if win._annotation_save_dir is not None
+            else None
+        )
+
         detection_results = _future.waitFor(
             detect_neurons(
                 working_stack,  # Prepared based on mock/real and single/multi
@@ -177,6 +192,7 @@ class CellDetector:
                 trim_edges=True,
                 min_volume_m3=win.ui.minVolumeSpin.value(),
                 n=None,
+                save_prefix=save_prefix,
             ),
             timeout=600,
         ).getResult()
@@ -233,7 +249,8 @@ class CellDetector:
                 filter=False,  # No extra filtering
                 transpose_display=True,  # stack is (n_frames, Y, X) after .T; row-major matches camera module orientation
                 custom_buttons=[("Center Camera", _center_camera_on_cell)],
-                save_dir=_annotation_save_dir(win.module.manager),
+                save_dir=getattr(win, "_annotation_save_dir", None),
+                base_name=getattr(win, "_annotation_base_name", None),
             )
 
             # from acq4_automation.object_detection import NeuronBoxViewer
--- a/acq4/modules/AutomationDebug/detection.py
+++ b/acq4/modules/AutomationDebug/detection.py
@@ -156,11 +156,15 @@ class CellDetector:
             "in_memory_stack_%Y%m%d_%H%M%S_%f"
         )
         annotation_save_dir = win.annotation_save_dir
-        save_prefix = (
-            str(annotation_save_dir / win.annotation_base_name)
-            if annotation_save_dir is not None
-            else None
-        )
+        if annotation_save_dir is None:
+            logger.warning(
+                "No annotation save directory available (set misc/cellAnnotationDir, "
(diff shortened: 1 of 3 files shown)


config/example/default.cfg
```diff
@@ -102,6 +102,12 @@ misc:
     ## organize their data.
     # storageDir: '/home/user/data'
 
+    ## Directory where cell health annotations (from the AutomationDebug "Detect
+    ## Neurons" workflow) are saved. Each annotation session is written to a
+    ## uniquely-named file here. If unset, defaults to a 'cell_annotations'
+    ## subdirectory of the data storage directory.
```
outofculture: reference `storageDir` in this, since it's the immediately previous setting

acq4/modules/AutomationDebug/detection.py
```diff
@@ -147,6 +165,20 @@ def _detectNeuronsZStack(
 
         win.cameraDevice.setFocusDepth(depth, name=f"{win.cameraDevice.name()} restore focus after detection z-stack")  # Restore focus
 
+        # Persist this detection session under one base name so the saved z-stack,
+        # cellpose masks, and annotations share the annotation tool's naming scheme
+        # and reload together. The base is shared with the annotation tool launched
+        # in _handleDetectResults.
+        win._annotation_save_dir = _annotation_save_dir(man)
+        win._annotation_base_name = datetime.datetime.now().strftime(
```
outofculture: these attrs are being accessed outside of self; they're not private. make them first class attrs or properties, as makes sense (does the `win` have a reference to `man`? if so, the whole `_annotation_save_dir` implementation could move into a property method).

acq4/modules/AutomationDebug/detection.py
```diff
@@ -216,6 +249,8 @@ def _center_camera_on_cell(context, _win=win):
                 filter=False,  # No extra filtering
                 transpose_display=True,  # stack is (n_frames, Y, X) after .T; row-major matches camera module orientation
                 custom_buttons=[("Center Camera", _center_camera_on_cell)],
+                save_dir=getattr(win, "_annotation_save_dir", None),
+                base_name=getattr(win, "_annotation_base_name", None),
```
outofculture: once these attrs are made proper, these getattrs can switch to normal attr access

config/example/default.cfg
outofculture: Done in aa0d51cf3 — the comment now refers to `storageDir` by name.

acq4/modules/AutomationDebug/detection.py
outofculture: Done in aa0d51cf3. `_annotation_save_dir` moved into an `AutomationDebugWindow.annotation_save_dir` property (the window reaches the manager via `self.module.manager`), and `annotation_base_name` is now a declared public attribute set when a detection run starts.

acq4/modules/AutomationDebug/detection.py
outofculture: Done in aa0d51cf3 — now plain attribute access (`win.annotation_save_dir` / `win.annotation_base_name`), no getattr.

acq4/modules/AutomationDebug/detection.py
```diff
@@ -147,6 +148,20 @@ def _detectNeuronsZStack(
...
         win.cameraDevice.setFocusDepth(depth, name=f"{win.cameraDevice.name()} restore focus after detection z-stack")  # Restore focus
 
+        # Persist this detection session under one base name so the saved z-stack,
+        # cellpose masks, and annotations share the annotation tool's naming scheme
+        # and reload together. The base is shared with the annotation tool launched
+        # in _handleDetectResults.
+        win.annotation_base_name = datetime.datetime.now().strftime(
+            "in_memory_stack_%Y%m%d_%H%M%S_%f"
+        )
+        annotation_save_dir = win.annotation_save_dir
```
outofculture, line 158: this value can be None; handle that situation gracefully

outofculture: Done in c80f5d068 — when `annotation_save_dir` is None (no `misc/cellAnnotationDir` and no storage dir), it now logs a clear warning that the z-stack, masks, and annotations won't be saved alongside the data, instead of silently falling back to the working directory.

